PR #1040, opened 2026-07-02
fix(testing): reap leaked WebView2 hosts on Windows app teardown (Closes #1022)

### LINKED ISSUE #1022: Bridge harness: WebView2 child processes leak on Windows app teardown, destabilizing back-to-back runs
## Summary

On Windows, each launched app instance spawns several `msedgewebview2.exe` child processes that are **not reaped** by `AppInstance.cleanup()` / `_terminate_tree`. They accumulate across test runs (observed **52** lingering `msedgewebview2.exe` after ~15 app launches in one session). Once enough pile up, later app instances get slow and unstable: runtimes climb (≈65s → ≈110s for the same suite) and tests intermittently **error in teardown** or fail on cold-start timeouts.

Discovered while running `tests/system/tests/test_windows_shells.py` for #975 (PR #1011). Symptoms:

- From a **clean** process state the suite is green and fast (11 passed, ~60s).
- On back-to-back runs without cleanup, the same suite degrades — a `teardown` ERROR appears on the last test (call phases still pass) and runtimes balloon.
- Manually killing stray `msedgewebview2.exe` (`taskkill /F /IM msedgewebview2.exe /T`) restores stability.

## Why it matters

Windows bridge CI (#804) runs many suites, each launching one or more app instances. Leaked WebView2 processes accumulate across the whole session, so the Windows lane would get progressively slower and flakier — exactly the platform the bridge port is meant to unlock.

## Likely cause

`_terminate_tree` (in `tests/system/termihub_harness/orchestrator.py`) kills the `termihub.exe` process tree, but WebView2 spawns its renderer/GPU/utility processes under a separate host (`msedgewebview2.exe`) that is not always a descendant of the killed PID, so they survive.

## Suggested fix

On Windows teardown, also reap the WebView2 children — e.g. after killing the app tree, enumerate and terminate `msedgewebview2.exe` processes whose user-data dir / parent maps to the just-killed instance (the app uses a per-instance `TERMIHUB_CONFIG_DIR`; the WebView2 user-data folder can be pinned per instance to make this precise), or kill WebView2 children spawned after the app launched. Best handled alongside the harness-resilience work in #1019.

armaxri:
## Summary

On Windows, each launched app instance spawns several `msedgewebview2.exe` renderer/GPU/utility processes under a host that is **not** a descendant of `termihub.exe`. `_terminate_tree` therefore missed them and they leaked across test runs (observed ~52 after ~15 launches), progressively slowing and destabilizing later instances until teardown ERRORs and cold-start timeouts appeared. Fixes #1022.

## Fix

- `AppInstance.start()` pins each instance's WebView2 user-data folder via `WEBVIEW2_USER_DATA_FOLDER` (Windows only). The app sets no Tauri `data_directory`, so wry passes an empty string to `CreateCoreWebView2EnvironmentWithOptions` and WebView2 honors the env var.
- `AppInstance.stop()` enumerates `msedgewebview2.exe` processes whose `--user-data-dir` matches this instance and terminates them (escalating to kill). Matching on the per-instance folder keeps it safe under parallel instances.
- Every teardown path (`restart`, `cleanup`, `__exit__`) funnels through `stop()`, so there is a single wiring point.

## Quality pass (`/simplify`)

- Extracted the shared terminate→wait→kill escalation into `_terminate_procs`, reused by `_terminate_tree` and the reaper.
- Name pre-filter before `proc.cmdline()` (costly on Windows) so the command line isn't read for the whole process table on every teardown — regression-tested.
- Inlined a single-use helper and dropped a redundant guard.

## Tests

TDD: the test commit precedes the fix, and all 8 new tests were confirmed to fail without the implementation. Coverage (cross-platform, via fakes — no live app launch):

- `_is_webview2_for_data_dir` match predicate: positive, wrong process name, other instance's dir, case/separator normalization.
- `_terminate_webview2_children`: reaps only matching hosts, escalates survivors to kill, no-ops off Windows, and skips the `cmdline()` read for non-WebView2 processes.
- `AppInstance.start()` pins `WEBVIEW2_USER_DATA_FOLDER` on Windows and omits it elsewhere.

`14 passed` (orchestrator) and `150 passed` (full non-integration suite).

## Test plan

- [x] `./tests/system/pytest.sh -m "not integration"` — 150 passed
- [x] Manual (Windows) — verified locally, see verification comment below: run `tests/system/tests/test_windows_shells.py` back-to-back several times and confirm no `msedgewebview2.exe` accumulation (`tasklist /FI "IMAGENAME eq msedgewebview2.exe"`) and stable runtimes.

### DIFF
--- a/tests/system/README.md
+++ b/tests/system/README.md
@@ -468,7 +468,10 @@ the old WebdriverIO find-by-title lookups. Tab lookups use the `TabsUi` mixin's
 ## Orchestration
 
 - `AppInstance(config_dir?)` — `start(bridge_port)`, `stop()`, `restart()`. Config
-  dir is stable across a restart so the saved last-session survives.
+  dir is stable across a restart so the saved last-session survives. On Windows
+  each instance pins its WebView2 user-data folder (via `WEBVIEW2_USER_DATA_FOLDER`)
+  so `stop()` can reap the `msedgewebview2.exe` hosts, which live outside the app
+  process tree and would otherwise leak across back-to-back runs (issue #1022).
 - `AgentInstance(host?, port?)` — `start()`, `stop()`, `restart()` for a
   `termihub-agent --listen` process, with process-tree teardown via `psutil`.
 
--- a/tests/system/termihub_harness/orchestrator.py
+++ b/tests/system/termihub_harness/orchestrator.py
@@ -127,6 +127,68 @@ def _terminate_tree(process: subprocess.Popen, timeout: float = 5.0) -> None:
             pass
 
 
+def _webview2_user_data_dir(config_dir: Path) -> Path:
+    """Per-instance WebView2 user-data folder, nested under the app config dir.
+
+    Passed to the app via ``WEBVIEW2_USER_DATA_FOLDER`` (Windows only) so every
+    launched instance pins its WebView2 host processes to a distinct folder —
+    which teardown then matches on to reap only that instance's children.
+    """
+    return config_dir / "webview2-user-data"
+
+
+def _is_webview2_for_data_dir(name: str, cmdline: list[str], user_data_dir: Path) -> bool:
+    """True if a process is a ``msedgewebview2.exe`` pinned to ``user_data_dir``.
+
+    WebView2 hosts its renderer/GPU/utility processes under
+    ``msedgewebview2.exe``. Those are launched with ``--user-data-dir=<folder>``
+    matching the app's ``WEBVIEW2_USER_DATA_FOLDER``, so matching on the folder
+    reaps exactly one instance's children — safe under parallel instances.
+    Comparison normalises path separators and case (Windows paths).
+    """
+    if (name or "").casefold() != "msedgewebview2.exe":
+        return False
+    needle = str(user_data_dir).replace("/", "\\").casefold()
+    joined = " ".join(cmdline or []).replace("/", "\\").casefold()
+    return needle in joined
+
+
+def _terminate_webview2_children(user_data_dir: Path, timeout: float = 5.0) -> None:
+    """Windows-only: reap the ``msedgewebview2.exe`` hosts for ``user_data_dir``.
+
+    WebView2's child processes live under a separate host that is *not* a
+    descendant of the app PID, so :func:`_terminate_tree` leaves them running.
+    They accumulate across app launches and eventually destabilise back-to-back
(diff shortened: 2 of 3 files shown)


armaxri:
## Verification (local Windows, debug build)

Ran `tests/system/tests/test_windows_shells.py` back-to-back with the fix, and probed the teardown moment directly to confirm the mechanism.

### Back-to-back suite runs (with the fix)

| Run | `msedgewebview2.exe` before → after | Result |
|-----|--------------------------------------|--------|
| 1 | 12 → 12 | 11 passed, 26.6s |
| 2 | 12 → 12 | 11 passed, 16.2s |
| 3 | 12 → 12 | 11 passed, 14.7s |

Flat count, stable/improving runtimes, no teardown ERRORs.

### Direct teardown probe

Launching one real app instance and measuring pinned hosts at teardown:

- **Env pin works** — 6 `msedgewebview2.exe` hosts spawned under this instance's `…\webview2-user-data` folder.
- **Matching is precise** — the reaper matched exactly those 6 and left the 12 unrelated machine-wide WebView2 processes untouched (a blanket `taskkill /IM msedgewebview2.exe` would have been wrong).
- **Reap clears them** — 0 remaining after `_terminate_webview2_children`.

### Why the leak didn't reproduce on this machine (and why that's fine)

Parent-chain probe during boot:

```
termihub.exe pid=22152
  pid=10788  ppid=22152 (termihub.exe)   --type=browser            <- in-tree
  pid=22120  ppid=10788 (msedgewebview2)  --type=renderer           <- in-tree
  pid=9868   ppid=10788 (msedgewebview2)  --type=gpu-process        <- in-tree
  pid=7020 / 19972       (msedgewebview2)  --type=utility            <- in-tree
  pid=18640              (msedgewebview2)  --type=crashpad-handler   <- in-tree
```

On **this** WebView2 runtime the browser host is a **direct child of `termihub.exe`**, so all helpers sit inside the app's process tree and `_terminate_tree` already reaps them — hence no leak here. Issue #1022 was observed on an environment where WebView2 launches the host **out-of-tree** (52 stragglers after ~15 launches), which is exactly what this reaper covers.

**Conclusion:** pin → match → reap all verified with no collateral damage and no regression. On an in-tree WebView2 build this is a harmless safety net; on the out-of-tree environments #1022 documented, it is the actual fix.

Manual test-plan item updated accordingly.

Issue #1240, opened 2026-07-07
Tunnel S1: surface Error in sidebar + Open Connections — Retry / View-last-error / force-Stop (GAP 3, frontend)


armaxri:
Part of the **SSH Tunnel Lifecycle Redesign** (concept-first parent #1191).
- Concept (source of truth): [`docs/concepts/backlog/ssh-tunnel-lifecycle-redesign.html`](../blob/develop/docs/concepts/backlog/ssh-tunnel-lifecycle-redesign.html)
- Audit: [`docs/audits/ssh-tunnel-state-machine.md`](../blob/develop/docs/audits/ssh-tunnel-state-machine.md)

**Stage S1 (GAP 3 UI + missing controls).** Depends on https://github.com/armaxri/termiHub/issues/1238. Once `get_statuses()` returns a persisted `Error`, surface it as the fourth resting colour (red) and add the controls the correct machine needs at rest: **Retry**, **View last error**, and **force-Stop** of an errored tunnel from the canonical kill panel.

### Scope
- [ ] **GAP 3 / persistence** — `loadTunnels` already overwrites `tunnelStates` from `get_statuses` (`src/store/appStore.ts:3541-3549`); confirm no store change is needed now that `Error` is returned, and that `error` survives a reload.
- [ ] **Kill panel filter** — Widen the Open Connections `activeTunnels` filter to include `error` (and stale `connected`): `src/components/OpenConnections/OpenConnectionsModal.tsx:117-120` / `:127-130`. Render a Stop row for errored tunnels; force-Stop calls `stop_tunnel`, which (per #1238) also clears the persisted error.
- [ ] **Sidebar row** — `TunnelListItem` already renders `status === "error"` text (`src/components/TunnelSidebar/TunnelListItem.tsx:145-147`); add the **Retry** (`Error → Connecting`, clears the stored error via `start_tunnel`) and **View last error** affordances, replacing the Play button for an errored row. Add the red `state-dot--error` resting dot (concept adds a `--color-error` variant).
- [ ] **Feedback rule** — Retry/force-Stop use the async Button + toast pattern (pending → success/error); no silent mutations.
- [ ] Compose from `src/components/ui/` primitives; delegate the visual/interaction pass to the `ui-design` subagent (design-system tokens only, no magic values).

### Done when
[...]

PR #1282, opened 2026-07-08
feat(tunnel): migrate TunnelListItem actions to shared Button primitive (#1259)

### LINKED ISSUE #1259: Migrate TunnelListItem icon-button group to shared Button primitive (per-button async feedback)
Follow-up to #1240 (PR to develop).

#1240 added Retry / View-last-error buttons to `TunnelListItem` but kept them as bare icon buttons to match the pre-existing sibling buttons in the same row (Start/Stop/Edit/Duplicate/Delete). Migrating only 2 of 7 would have made the row visually inconsistent, so the whole group is deferred as one unit.

### Scope
- [ ] Migrate the entire `.tunnel-item__actions` icon-button group in `src/components/TunnelSidebar/TunnelListItem.tsx` (Start, Stop, Retry, View-last-error, Edit, Duplicate, Delete) to the shared `Button` primitive (`src/components/ui/`).
- [ ] Async actions (Start/Stop/Retry) use the async Button pending state + toast feedback (pending → success/error), no silent mutations.
- [ ] Tokens only; delegate the visual/interaction pass to the `ui-design` subagent.
- [ ] Update/extend the `TunnelListItem` tests accordingly.

### Done when
Every action button in a tunnel row is a shared `Button` primitive, async actions show per-button pending feedback, and the row remains visually consistent — with tests green.

armaxri:
Closes #1259 (follow-up to #1240).

## Approach

Migrates the entire `.tunnel-item__actions` icon-button group in
`TunnelListItem` (Start, Stop, Retry, View last error, Edit, Duplicate, Delete)
from bespoke `.tunnel-item__action` buttons to the shared ghost `Button`
primitive, matching the merged NetworkTools sidebar precedent.

- **Real async feedback**: Start / Stop / Retry return the store's
  `startTunnel` / `stopTunnel` promise from `onClick`, so the Button drives a
  per-button pending spinner (disabled + `aria-busy`) tied to the actual backend
  call and flashes success on completion. The store already emits its own
  loading→success/error toast and re-throws, so `errorToast={false}` avoids a
  duplicate error toast while the Button still returns to idle on rejection.
- **Prop types**: `onStart` / `onStop` widened to `(id) => void | Promise<void>`
  so the async lifecycle is wired to the real calls, not cosmetic.
- **Tokens only**: removed the dead bespoke button CSS (color/hover/radius/size);
  all skinning now comes from `.ui-btn--ghost`. A single geometry-only override
  keeps the dense error-state row (up to five actions) compact.
- Stable `data-testid`s and `aria-label`s are preserved; the testid catalog is
  unchanged.

## Test plan

- New `TunnelListItem.button.test.tsx` (TDD, committed before the impl): asserts
  the actions render as `ui-btn ui-btn--ghost` primitives and that Start / Stop /
  Retry enter `ui-btn--pending` while the promise is in flight and return to idle
  on resolve. Written red → green.
- `pnpm test` (TunnelSidebar suite: 13 passed), `pnpm build` (tsc), `pnpm run lint`,
  and `./scripts/check.sh` all pass. No `any` types.
- **Manual visual check (macOS, not automatable here)**: open the Tunnel sidebar,
  hover a row and confirm the ghost action buttons look consistent with the rest
  of the app; start a tunnel and confirm the per-button pending spinner appears
  and clears; trigger an error and confirm the Retry / View-last-error row stays
  visible and Retry shows a spinner.

### DIFF
--- a/docs/changes/dev2/feature/1259-tunnel-listitem-button-primitive.md
+++ b/docs/changes/dev2/feature/1259-tunnel-listitem-button-primitive.md
@@ -1,0 +1,8 @@
+### Changed
+
+- Tunnel sidebar rows now compose their action controls (Start, Stop, Retry,
+  View last error, Edit, Duplicate, Delete) from the shared `Button` primitive
+  instead of bespoke icon buttons. The async actions (Start / Stop / Retry) show
+  a per-button pending spinner tied to the real backend call and flash success on
+  completion, so the whole row is visually consistent with the rest of the app
+  and every action gives immediate feedback (#1259, follow-up to #1240).
--- a/src/components/TunnelSidebar/TunnelListItem.button.test.tsx
+++ b/src/components/TunnelSidebar/TunnelListItem.button.test.tsx
@@ -1,0 +1,189 @@
+/**
+ * Verifies the tunnel list-item action row is composed from the shared `Button`
+ * primitive (issue #1259, follow-up to #1240) rather than bespoke
+ * `.tunnel-item__action` buttons.
+ *
+ * Every action button must carry the `ui-btn` primitive classes (ghost variant),
+ * and the async actions (Start / Stop / Retry) must drive the Button's async
+ * lifecycle — a pending spinner state while the underlying promise is in flight,
+ * returning to idle once it resolves. The stable `data-testid`s survive.
+ */
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import { act } from "react";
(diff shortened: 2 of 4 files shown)


armaxri:
Follow-up filed: #1284 — add a first-class icon-only variant to the shared `Button` primitive so `TunnelListItem` (this PR) and the NetworkTools icon rows don't each need a component-local geometry override. Left unassigned.

Issue #1355, opened 2026-07-10
feat(agent): optional agent-side GitHub self-update (off by default)


armaxri:
**Part of Epic #1345 · Concept** (`docs/concepts/backlog/remote-agent-update-strategy.html`)

## Summary

Optional add-on / Approach 1: agent-side self-update — a 24 h GitHub-releases poll, SHA-256-verified download, and (when idle + deferred) auto-apply. Off by default, gated by `allow_self_update`.

## Scope (in)

- 24 h timer in the agent (`agent/src/main.rs` run loops), GitHub `releases/latest` poll, semver compare vs `env!("CARGO_PKG_VERSION")`.
- SHA-256-verified download (reuses SI-4).
- `agent.update_available` notification → self-update toast (dialogs mockup state 4); when no hosts connected and deferred is enabled, auto-download + defer (via SI-6).
- Agent outbound HTTP near `agent/src/network/mod.rs`; `last_check_time` in `state.json`; notification type in `protocol/methods.rs`. Gated by `allow_self_update` (SI-7).

## Out of scope

Any always-on behavior; air-gapped fallback beyond a graceful skip.

## Technical notes

- Must skip cleanly without internet (log a warning, no crash).

## Acceptance criteria

- With self-update enabled and a newer release, the agent notifies connected hosts (toast) and, when idle + deferred, applies it.
- Disabled by default; no network → logged warning, no crash.

## Test plan (TDD)

- **Unit:** release-JSON parse + semver-newer decision; verified-download gate (Rust, mocked HTTP).
- **Integration (Docker #995):** mock releases endpoint.
- **Manual:** enable on a test host, publish a newer tag, observe the toast.

armaxri:
## Investigation — partially complete, kept open

PR #1402 (merged) delivered the **poll → stage** half of this issue's scope, but the **auto-apply-on-idle** half of the acceptance criteria is intentionally NOT done.

**Done (via #1402):**
- 24h GitHub `releases/latest` poll + semver compare vs `CARGO_PKG_VERSION`
- SHA-256-verified download (fail-closed on mismatch/missing checksum)
- `agent.update_available` notification → desktop self-update toast
- Staging: records `update.pending_update` + `last_check_time` in `state.json`
- Gated behind `allow_self_update` (no task / zero outbound network when off); offline → logged warning, no crash

**Remaining (why this stays open):**
[...]

Issue #1345, opened 2026-07-10
[Epic] Remote agent update strategy (coordinated, visible updates)


armaxri:
Parent tracking issue for implementing a **coordinated, visible remote-agent update strategy**.

**Concept (source of truth):** `docs/concepts/backlog/remote-agent-update-strategy.html`

The concept is authoritative. Where code and concept disagree, fix the code — **except** where a real constraint has already moved the code ahead of the concept, in which case correct the concept (two such corrections are called out below).

## Background

termiHub deploys a `termihub-agent` binary to each remote host and talks to it over JSON-RPC (`docs/remote-protocol.md`). Today the desktop already probes the agent's version on connect (`src-tauri/src/utils/version.rs::check_version` → `VersionStatus::{Compatible, AgentTooOld, MajorMismatch, InvalidVersion}`; handshake constant `AGENT_PROTOCOL_VERSION` at `agent/src/handler/dispatch.rs:44`) and can force a hard redeploy — but there is **no coordination**: when one desktop updates a shared agent, other desktops' active sessions are silently killed, and no in-app surface communicates an agent's version, update state, or who else is connected. This epic delivers a coordinated, visible update strategy in dependency-ordered phases (guard → broadcast → deferred → optional self-update), each independently valuable.

## Key decisions (from the concept)

- **Compatibility rule (already exists, codify as policy):** same major **and** `agent.minor ≥ desktop.minor` (patch ignored). A newer agent always serves an older desktop — backward-compat is one-directional. This is the single source of truth for the four badge states.
- **Update triggers — three modes, phased:** (1) _manual/guard_ (Approach 2) — pre-check `agent.list_connections`; 0 others → update now, N others → warn + explicit `update_agent_force`; (2) _coordinated broadcast_ (Approach 3) — `agent.request_update` → `agent.update_pending` to other clients who disconnect cleanly and auto-reconnect, agent proceeds after all ack or 10 s timeout; (3) _deferred_ (Approach 4) — persist `pending_update`, apply only at zero active sessions; (+) _optional agent self-update_ (Approach 1) — off by default, 24 h GitHub poll, exec-replace when idle.
- **Redeploy while sessions may be live:** persistent sessions run as **detached daemon subprocesses** (`--daemon <id>`, `setsid`-detached, commit `7c52017`) that outlive the agent worker, so a binary swap does not kill them; the desktop reconnects and re-attaches. Approach 4 leans entirely on this.
[...]

Issue #995, opened 2026-06-30
Remote agent: live connect + child sessions E2E (needs a deployed-agent Docker fixture)


armaxri:
Follow-up from #974. The remote-agent port (`tests/system/tests/test_remote_agent.py`) covers create / context-menu / connection-error dialog / setup-wizard against the existing password-SSH container (no agent binary). The **success path** is still uncovered because it needs a remote host with the `termihub-agent` binary actually deployed:

- connect to a remote host running the agent (old `remote-agent` container, port 2211, compose profile `agent`),
- expand the agent folder and see available shells,
- create a shell session under the connected agent,
- reconnect and re-attach persistent sessions,
- the connect/disconnect/new-session/edit/delete context menu against a *connected* agent.

These were the `pending` describes in the old `infrastructure/remote-agent.test.js`. Implementing them needs a Docker fixture that builds/ships the agent binary into the container (cross-compiled to the container arch) and a conftest fixture to bring it up.

The `agent-missing` "Setup Agent" button (old MT-AGENT-02) is **not** part of this — it is covered by the `classifyAgentError` / `ConnectionErrorDialog` unit tests (connecting to a no-agent host classifies as generic "Connection Failed", not `agent-missing`).

armaxri:
Implemented in #1042 (branch `feature/995-remote-agent-live-e2e`, targeting `develop`).

The deployed-agent Docker fixture + live E2E cover all four `pending` describes: live connect + available shells, a child shell session, persistent-session reconnect/re-attach, and the connect/disconnect/new-session/edit/delete menu on a connected agent.

While wiring up the reconnect test, it surfaced a real bug — persistent agent sessions were killed on every SSH disconnect (SIGHUP from the SSH login session + the daemon inheriting the SSH channel's stderr). #1042 also fixes that by fully detaching the daemon on unix (`setsid` + stderr → per-session log file), mirroring the existing Windows detachment.

PR #1518, opened 2026-07-14
feat(agent): deferred update (apply on last disconnect)

### LINKED ISSUE #1352: feat(agent): deferred update (apply on last disconnect)
**Part of Epic #1345 · Concept** (`docs/concepts/backlog/remote-agent-update-strategy.html`)

## Summary

Phase 3 / Approach 4: a deferred update that leaves active sessions untouched and applies only when the last session disconnects — leaning on the detached-daemon model so persistent sessions survive the binary swap.

## Scope (in)

- RPC `agent.request_deferred_update { binary_path }`; persist `pending_update` in `state.json` (`agent/src/state/persistence.rs::AgentState`).
- Zero-session callback in `agent/src/session/manager.rs` (`active_count()==0` hook) execs the new binary via the daemon detach model (`agent/src/daemon/process.rs`, commit `7c52017`).
- Desktop deferred-update banner with "Apply Now" (dialogs mockup state 3).

## Out of scope

Self-update; GitHub polling.

## Technical notes

- Note the concept caveat: the update may be delayed indefinitely with long-running persistent sessions. Banner via **`ui-design`**. Gate Unix-only exec-replace with `#[cfg(unix)]`.

## Acceptance criteria

- Requesting deferred update leaves active sessions untouched; on last disconnect the agent applies the update; next connect reports the new version; "Apply Now" forces it immediately.

## Test plan (TDD)

- **Unit:** `pending_update` round-trips through `state.json`; the zero-session callback fires exactly at count 0 (Rust).
- **Integration (Docker #995):** open a session, request deferred, close the session, assert the new version on reconnect.
- **System (Python bridge):** banner + Apply Now.
- **Manual:** long-running session, verify no interruption.

armaxri:
Closes #1352
Part of Epic #1345

## Summary

Phase 3 / Approach 4: a **deferred agent update** that leaves active sessions untouched and applies only when the **last session disconnects**, leaning on the detached-daemon model so persistent sessions survive the binary swap and report the new version on the next connect.

- **RPC `agent.request_deferred_update { binaryPath?, version? }`** — records a `pending_update` in `state.json`. Omit `binaryPath` to apply an already-staged self-update ("Apply Now"). Applies **immediately only when idle** (0 active sessions); otherwise defers.
- **State persistence** — reuses the existing `UpdateState.pending_update` in `state.json` (from #1355); `SessionManager` now routes state through a configurable path.
- **Zero-session hook** — `SessionManager::close` computes the remaining running count and, on reaching zero with a pending update, applies it. A pure `should_apply_deferred_update(active, has_pending)` gate is the single source of truth (true only at 0 sessions + pending).
- **Exec-replace via daemon detach** — apply swaps the on-disk agent binary with the staged one (atomic temp-file + rename, marked executable) and re-execs with the same args. Gated `#[cfg(unix)]`; non-Unix returns an error and keeps compiling.
- **Desktop banner** — a per-agent sidebar banner with **Apply Now** (idle → applies + reconnects; busy → defers, naming the active-session count) and **Dismiss**, composed from the shared `Button` primitive with toast feedback. The desktop backend forwards the agent's `agent.update_available` notification as an `agent-update-available` event.

## How the zero-session apply works

`request_deferred_update` persists the pending update, then checks `active_count()`. If **0**, it consumes the pending update (clearing it from `state.json` first, so a re-exec can never loop) and calls the injected `UpdateApplier`. If **> 0**, it returns `Deferred { activeSessions }` and does nothing to running sessions. On each `close`, when the last running session is removed the same apply path fires. On Unix a successful apply re-execs (the connection drops — the desktop treats that as expected and reconnects to see the new version).

## Exec-replace gating

`apply_update_binary` is `#[cfg(unix)]` (copy staged → temp in the target dir, `chmod 0755`, atomic `rename`, then `CommandExt::exec`). `#[cfg(not(unix))]` returns an error (`-32016`), so Windows/other builds compile — CI validates the non-Unix build.

## Constraint honored

Active sessions are **never** interrupted: apply is strictly gated on `active_count == 0`. "Apply Now" with active sessions defers rather than forcing (the banner reports the count).

## Documented caveat

With a long-running persistent session the update may be delayed **indefinitely** — it only applies once the last session disconnects. This is by design.

## Tested vs deferred

**Unit-tested (runnable on macOS):**
- `should_apply_deferred_update` gate — applies only at 0 sessions with a pending update.
- Unix binary-swap helper — replaces contents + marks executable, no temp left behind.
- `state.json` round-trip for a staged pending update (and legacy state without the field → `None`).
- Zero-session hook via an injected recording `UpdateApplier` + isolated state path: applies exactly on last disconnect; not before; not without a pending update.
- `request_deferred_update`: applies when idle, defers with active sessions, applies existing pending when no path given, and errors on missing binary / no pending.
- Frontend: `AgentUpdateBanner` Vitest suite (renders, Apply Now applied/deferred, Dismiss).

**Deferred (not runnable here):**
- Docker integration test (#995): open a session → request deferred → close → assert new version on reconnect. Follow-up filed.
- Python-bridge banner E2E. Follow-up filed.

## Out of scope

Self-update and GitHub polling (from #1355) are untouched; auto-triggering the background self-update timer's apply-on-idle onto this mechanism is left as a tracked follow-up.

## Test plan

- `cargo test -p termihub-agent` (311 unit + integration green), `cargo clippy -p termihub-agent --all-targets -- -D warnings` clean.
- `cargo build -p termihub` + `cargo clippy -p termihub` clean.
- `pnpm exec tsc --noEmit`, `pnpm run lint`, `pnpm run format:check`, `pnpm exec vitest run src/components/AgentUpdateBanner` green; `pnpm run markdownlint` clean.
- Manual steps added to `docs/testing.md` (deferred update: long-running session → request → no interruption; close last session → applies; reconnect → new version; Apply Now forces it when idle).

### DIFF
--- a/agent/src/handler/dispatch.rs
+++ b/agent/src/handler/dispatch.rs
@@ -25,20 +25,23 @@ use crate::monitoring::MonitoringManagerApi;
 use crate::network;
 use crate::protocol::errors;
 use crate::protocol::methods::{
-    AgentSettings, AgentSettingsUpdateParams, AgentShutdownParams, AgentShutdownResult,
-    Capabilities, ConnectionCreateParams, ConnectionDeleteParams, ConnectionInfo,
-    ConnectionListResult, ConnectionTypesResult, ConnectionUpdateParams, FilesDeleteParams,
-    FilesListParams, FilesListResult, FilesMkdirParams, FilesReadParams, FilesReadResult,
-    FilesRenameParams, FilesStatParams, FilesWriteParams, FolderCreateParams, FolderDeleteParams,
-    FolderUpdateParams, HealthCheckResult, InitializeParams, InitializeResult,
-    MonitoringSubscribeParams, MonitoringUnsubscribeParams, NetworkDnsLookupParams,
-    NetworkPingParams, NetworkPortScanParams, NetworkTracerouteParams, NetworkWolParams,
-    SessionAttachParams, SessionCloseParams, SessionCreateParams, SessionCreateResult,
-    SessionDetachParams, SessionGetBufferParams, SessionGetBufferResult, SessionInputParams,
-    SessionListEntry, SessionListResult, SessionResizeParams,
+    AgentRequestDeferredUpdateParams, AgentRequestDeferredUpdateResult, AgentSettings,
+    AgentSettingsUpdateParams, AgentShutdownParams, AgentShutdownResult, Capabilities,
+    ConnectionCreateParams, ConnectionDeleteParams, ConnectionInfo, ConnectionListResult,
+    ConnectionTypesResult, ConnectionUpdateParams, FilesDeleteParams, FilesListParams,
+    FilesListResult, FilesMkdirParams, FilesReadParams, FilesReadResult, FilesRenameParams,
+    FilesStatParams, FilesWriteParams, FolderCreateParams, FolderDeleteParams, FolderUpdateParams,
+    HealthCheckResult, InitializeParams, InitializeResult, MonitoringSubscribeParams,
+    MonitoringUnsubscribeParams, NetworkDnsLookupParams, NetworkPingParams, NetworkPortScanParams,
+    NetworkTracerouteParams, NetworkWolParams, SessionAttachParams, SessionCloseParams,
+    SessionCreateParams, SessionCreateResult, SessionDetachParams, SessionGetBufferParams,
+    SessionGetBufferResult, SessionInputParams, SessionListEntry, SessionListResult,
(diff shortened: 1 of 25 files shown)


armaxri:
Follow-up issues for deferred work referenced above:
- #1519 — Docker (#995) integration test for the deferred-update cycle
- #1520 — Python-bridge E2E for the deferred-update banner
- #1521 — auto-apply staged self-update on idle via this mechanism (completes the #1355 deferral)

Issue #1519, opened 2026-07-14
test(agent): Docker integration test for deferred agent update


armaxri:
Follow-up to #1352 (PR #1518).

The deferred-update apply-on-last-disconnect flow is covered by agent unit tests (injected `UpdateApplier` + isolated state path), but the real binary swap + re-exec over a live agent is not exercised end-to-end.

## Scope
- [ ] Using the #995 live-agent Docker harness, open a session on a deployed agent, call `agent.request_deferred_update` with a staged newer binary, then close the session.
- [ ] Assert the agent swaps + re-execs on last disconnect and reports the new version on reconnect.
- [ ] Assert an active session is NOT interrupted while the update is pending.

## Done when
A Docker-backed integration test drives the full deferred-update cycle and passes in CI (Linux).

PR #1533, opened 2026-07-14
feat(agent): auto-apply staged self-update on idle (Closes #1401)

### LINKED ISSUE #1401: Auto-apply staged agent self-update on idle (wire to SI-6 / #1352)
Follow-up to #1355 (PR pending). Part of epic #1345.

#1355 landed the agent-side self-update **poll → semver → SHA-256-verified download → notify/stage** pipeline (gated behind \`allow_self_update\`, off by default). When the agent is idle it downloads and verifies a newer binary, records it as \`update.pending_update\` in \`state.json\`, and **stops there** — the actual apply-on-idle step is intentionally deferred because it depends on the deferred-apply / exec-replace mechanism from **SI-6 (#1352)**, which is not yet built.

This issue tracks wiring the apply step once #1352 lands.

## Scope
- [ ] When a `pending_update` is recorded and the deferred-apply mechanism (#1352) is available, apply the staged, verified binary on reaching zero active sessions (exec-replace via the daemon's survive-restart model).
- [ ] Clear `update.pending_update` from `state.json` after a successful apply; keep it (and log) on failure so a later cycle can retry.
- [ ] Re-emit / update the `agent.update_available` notification lifecycle so connected desktops see the applied version after restart.
- [ ] Respect the connection's `update_strategy` (immediate / coordinated / deferred) when choosing whether to auto-apply.
- [ ] Add tests for the apply-on-idle transition and the state-clearing behavior.

## Done when
An agent with `allow_self_update` on that has staged a verified newer binary automatically applies it once its last session closes (via #1352's mechanism), comes back on the new version, and clears `pending_update` — with no forced disconnects.

armaxri:
Closes #1401
Part of Epic #1345
Follow-up to #1355 (PR #1402) + #1352 (PR #1518)

## Summary

Wires the agent-side self-update **staging** pipeline (#1355) through the **deferred-apply / exec-replace** mechanism (#1352) so a staged, verified update actually applies once the agent goes idle — without ever interrupting active sessions.

### What was already wired (by #1352)
- The zero-session `SessionManager::close()` hook already applied *any* recorded `pending_update` on last disconnect via `apply_pending_update` → `UpdateApplier` (unix exec-replace), plus the pure `should_apply_deferred_update` gate and `agent.request_deferred_update`.

### What this PR adds
- **Bridged the two mechanisms (single source of truth).** The #1355 timer wrote `pending_update` directly to `state.json`, while `SessionManager` owns a *cached* in-memory `AgentState` — so a staged self-update was invisible to the apply path and got clobbered on the next `close()`. All `update` state (last-check time + staged pending update) now flows through `SessionManager` (`record_update_check_time`, `stage_pending_update`, `request_deferred_update`), so both paths share one state.
- **Auto-apply on idle.** After staging while idle, the timer hands the verified binary to `request_deferred_update`, which applies immediately (exec-replace) — or, if a session raced in, re-checks and defers to the last-disconnect hook. Active sessions are never cut.
- **`update_strategy` gate.** The agent now receives the connection's configured strategy via a new `--update-strategy` CLI flag. `immediate`/`deferred` auto-apply on idle; `coordinated` stages + notifies only.
- **`allow_self_update` gate.** Implied — the self-update task is never spawned when off.
- **Retain-on-failure.** `apply_pending_update` now applies **first** and clears the pending update only on **success**; a failed apply keeps it (and logs) so a later cycle retries. (Previously it cleared before applying, losing the update on failure.)
- **Notification re-emit.** `agent.update_available` is emitted **before** the apply so connected desktops observe the lifecycle even though a successful unix apply re-execs and never returns.

### Never-interrupt guarantee
Auto-apply fires strictly at `active_count() == 0` (reusing #1352's `should_apply_deferred_update`); persistent daemon sessions survive the exec-replace and re-attach.

## Testing
Unit tests (macOS-runnable):
- `agent/src/update/mod.rs`: apply-on-idle for eligible strategy (applies + clears pending); coordinated stages without applying; failed apply keeps pending; active-sessions notify-only; offline/up-to-date.
- `agent/src/session/manager.rs`: successful apply on last disconnect clears pending; failed apply retains it.
- `agent/src/main.rs`: `--update-strategy` parsing.
- `src-tauri/src/terminal/backend.rs`: desktop exec command carries `--update-strategy`.

Deferred: Docker (#995) end-to-end apply-across-restart integration is tracked by the existing deferred-update Docker follow-up **#1519** (stage → close last session → assert new version on reconnect). Manual steps added to `docs/testing.md`.

## Test plan
- [x] `cargo test -p termihub-agent` (317 unit + integration pass)
- [x] `cargo test -p termihub` backend tests
- [x] `cargo clippy -p termihub-agent -p termihub --all-targets -- -D warnings`
- [x] `cargo fmt --check`
- Manual: enable self-update on a Linux host, publish a newer tag, keep a session open → no apply; close last session → applies; reconnect → new version; `pending_update` cleared (see docs/testing.md).

### DIFF
--- a/agent/src/io/stdio.rs
+++ b/agent/src/io/stdio.rs
@@ -20,6 +20,7 @@ use crate::session::manager::SessionManager;
 pub async fn run_stdio_loop(
     shutdown: CancellationToken,
     allow_self_update: bool,
+    update_strategy: crate::update::UpdateStrategy,
 ) -> anyhow::Result<()> {
     let (notification_tx, mut notification_rx) =
         tokio::sync::mpsc::unbounded_channel::<JsonRpcNotification>();
@@ -41,7 +42,11 @@ pub async fn run_stdio_loop(
 
     // Optional background GitHub self-update check (off unless opted in).
     crate::update::spawn_self_update_task(
-        crate::update::UpdateConfig::from_env(allow_self_update, env!("CARGO_PKG_VERSION")),
+        crate::update::UpdateConfig::from_env(
+            allow_self_update,
+            update_strategy,
+            env!("CARGO_PKG_VERSION"),
+        ),
         session_manager.clone(),
         update_tx,
         shutdown.child_token(),
--- a/agent/src/io/tcp.rs
+++ b/agent/src/io/tcp.rs
@@ -25,6 +25,7 @@ pub async fn run_tcp_listener(
     addr: &str,
     shutdown: CancellationToken,
     allow_self_update: bool,
+    update_strategy: crate::update::UpdateStrategy,
 ) -> anyhow::Result<()> {
     let listener = TcpListener::bind(addr).await?;
     info!("Listening on {}", listener.local_addr()?);
@@ -48,7 +49,11 @@ pub async fn run_tcp_listener(
 
     // Optional background GitHub self-update check (off unless opted in).
     crate::update::spawn_self_update_task(
-        crate::update::UpdateConfig::from_env(allow_self_update, env!("CARGO_PKG_VERSION")),
+        crate::update::UpdateConfig::from_env(
+            allow_self_update,
+            update_strategy,
+            env!("CARGO_PKG_VERSION"),
+        ),
         session_manager.clone(),
         update_tx,
         shutdown.child_token(),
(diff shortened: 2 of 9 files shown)


armaxri:
Deferred end-to-end coverage tracked in sibling follow-up #1534 (self-update poll → stage → auto-apply-on-idle over the #995 live-agent Docker harness), alongside #1519 (desktop-pushed `request_deferred_update` path).

PR #1550, opened 2026-07-16
test(agent): live integration test for self-update auto-apply on idle (Closes #1534)

### LINKED ISSUE #1534: test(agent): Docker integration test for self-update auto-apply on idle
Follow-up to #1401 (PR #1533). Sibling of #1519.

#1401 wired the agent's background self-update staging (#1355) through the deferred exec-replace mechanism (#1352) so a staged, verified update auto-applies on idle, gated by `update_strategy`. This is covered by agent unit tests (mock GitHub server + injected `UpdateApplier` + isolated state), but the real GitHub-poll → stage → binary-swap → re-exec cycle is not exercised end-to-end over a live agent. #1519 covers the *desktop-pushed* `agent.request_deferred_update` path; this issue covers the *self-driven* poll path.

## Scope
- [ ] Using the #995 live-agent Docker harness, run an agent with `--allow-self-update --update-strategy deferred` pointed at a mock/staged newer release so it downloads + SHA-256-verifies + stages a `pending_update`.
- [ ] With an active session open, confirm the agent does NOT swap its binary (session uninterrupted, `pending_update` retained).
- [ ] Close the last session; assert the agent exec-replaces, comes back on the new version on reconnect, and clears `pending_update` from `state.json`.
- [ ] Assert a `coordinated` strategy stages + notifies but does NOT auto-apply on idle.
- [ ] Assert a failed apply (e.g. corrupted staged binary) keeps `pending_update` for retry.

## Done when
A Docker integration test drives the self-update poll → stage → auto-apply-on-idle cycle end-to-end and asserts the new version after reconnect, the strategy gate, the never-interrupt guarantee, and the retain-on-failure behavior.

armaxri:
## What

Adds a live-agent integration test for the agent's self-update **auto-apply-on-idle** behaviour (#1401, follow-up to PR #1533; sibling of #1519).

The self-update logic is already unit-tested in `agent/src/update/mod.rs`, but those tests inject a **non-re-execing** `UpdateApplier` and an isolated `state.json` — so the *real* GitHub-poll → download → SHA-256-verify → binary-swap → **re-exec** cycle is never exercised against a running agent. This PR fills that gap.

`agent/tests/self_update_integration.rs` launches a real `termihub-agent --listen --allow-self-update --update-strategy <s>` child (from a throwaway copy of the built binary, so the self-replace doesn't clobber cargo's artifact) and points it at an in-process `wiremock` "mock GitHub" via new `TERMIHUB_AGENT_UPDATE_*` env seams.

### Cases
- **deferred + idle** → really swaps its binary (inode changes) and re-execs, then comes back alive and reconnectable.
- **coordinated + idle** → stages + records `pending_update`, but never swaps (no auto-apply on idle).
- **active session is never interrupted** → with a live **local shell** *and* a real **Docker container** session open, the poll neither stages nor swaps; the session survives.
- **failed apply** (read-only install dir) → keeps `pending_update` for retry, agent stays on its old binary.

Unix-only (the self-replace + re-exec is Unix-only); the Docker case skips cleanly when Docker is unavailable, so this runs on Linux CI.

## Why a live process, not the SSH/Docker `remote-agent` container (#995)?

The self-update mechanism is entirely process-level (poll a URL, swap the on-disk binary, `execve`) and needs neither SSH nor a container to be exercised faithfully — the piece under test is the *live agent process*, which a child `--listen` process reproduces exactly and deterministically. The one case that genuinely benefits from a container (never-interrupt against a real remote session) additionally runs against a live Docker session. Rationale is documented in the test's module doc.

## Production change

One small, test-only seam commit: `UpdateConfig::from_env` gains three env overrides (endpoint, asset suffix, first-poll delay), all unset/no-op in production. The background task now schedules its first tick via `interval_at`; with the default zero delay it still fires immediately, so shipping behaviour is unchanged.

## Follow-up filed

The end-to-end test surfaced that on the **real Unix path a successful apply does not clear `pending_update`**: `apply_pending_update` re-execs before the in-process clear runs, and nothing clears an already-applied update at startup — so the record persists across the re-exec (unlike the unit tests, whose injected applier lets the clear run). Filed as a separate follow-up issue; this test pins the real behaviour.

## Test plan
- `cargo test -p termihub-agent --test self_update_integration` — 5/5 pass locally (incl. the Docker case), stable across repeated parallel runs.
- `cargo test -p termihub-agent --bin termihub-agent update::` — 36 update unit tests still pass.
- `cargo clippy -p termihub-agent --all-targets --all-features`, `cargo fmt --all --check`, markdownlint — clean.


armaxri:
Follow-up issue filed: #1551 (successful deferred/self-update apply never clears `pending_update`, causing a spurious re-exec on the next idle). This PR's test currently pins the real behaviour; #1551 will flip that assertion.

Issue #1545, opened 2026-07-16
test-infra: test-mode bridge verb to inject Tauri events (emitEvent) for system tests


armaxri:
Enabler split out of #1520 (see that issue's investigation for full detail).

## Problem
The Python-bridge system-test dispatcher (`src/testbridge/dispatcher.ts`) is DOM-only: it has no way to emit a Tauri event or mutate the store. This blocks any system test that depends on a backend-originated event to drive UI.

Concrete blocker: the deferred-update banner (`src/components/AgentUpdateBanner/AgentUpdateBanner.tsx`) renders only when `agentUpdates[agentId]` is populated, and that store slice is fed **exclusively** by the `agent-update-available` Tauri event. Nothing in a system test can currently make that event fire.

## Proposal
Add a **test-mode-gated** bridge verb (TS dispatcher + Python client parity) to emit a named Tauri event with a payload, so tests can surface such UI via the real event path rather than reaching into the store. Gate it so it is inert outside the test harness.

## Scope / acceptance
- New verb wired through the TS dispatcher and the Python bridge client with matching signatures.
- A smoke test proving an injected `agent-update-available` event surfaces the banner.
- Docs note in `docs/testing.md` on the verb and its test-only gating.

Blocks #1520. Found while implementing #1520.

Issue #1531, opened 2026-07-14
feat(backend): emit remote path in transfer-progress for Transfer Queue rows


armaxri:
Follow-up to #1337 (PR #1530), part of Epic #1331.

The Transfer Queue panel (#1337) renders a remote-path cell per row, but the backend `transfer-progress` event payload carries no path field, so `TransferEntry.path` is always undefined and the cell is omitted. The concept mockup (`docs/concepts/backlog/ftp-client.html`) shows the remote path alongside the file name.

## Scope

- [ ] Add a `path` (remote path) field to the `TransferProgress` payload emitted by the SFTP (#1245) and FTP (#1336) transfer paths.
- [ ] Surface it in `src/services/api.ts` `TransferProgress` and map it into `TransferEntry.path` in `transferEntryFromProgress` (`src/types/transfer.ts`).
- [ ] The panel already renders `.transfer-row__path` when `path` is set — no UI change needed beyond the mapping.

## Done when

A live transfer row in the Transfer Queue panel shows the remote path (e.g. `/uploads/data.csv`) as in the mockup, covered by an updated mapper unit test.

Issue #1337, opened 2026-07-10
feat(ui): Transfer Queue panel (rows, controls, minimized state)


armaxri:
**Part of Epic #1331 · Concept #518** (`docs/concepts/backlog/ftp-client.html`)

## Summary

The connection-type-agnostic **Transfer Queue** panel: docked above the status bar with per-transfer rows, controls, and a minimized status-bar indicator, reflecting the SI-5 backend events.

## Scope (in)

- New `src/types/transfer.ts` (richer `TransferEntry`: `state`, `speedBytesPerSec`, `totalBytes|null`, `error`).
- Transfer slice in `src/store/appStore.ts` (`addTransfer`/`updateTransfer`/`removeTransfer`/`clearCompleted`).
- `src/components/TransferQueue/{TransferQueue,TransferEntry,TransferControls}.tsx` + panel CSS; dock above the status bar; per-row direction/name/path/progress/percent/throughput + Pause/Resume/Cancel/Retry/Remove; footer Clear Completed / Cancel All; **minimized** status-bar indicator with live count that re-expands.
- Wire `src/services/events.ts` (`transfer-progress`) and `src/services/api.ts`.

## Out of scope

Backend logic (SI-5).

## Technical notes

- **Delegate to the `ui-design` subagent** — compose from `src/components/ui/` primitives, tokens only (`src/styles/variables.css`), one motion language, feedback on every action. Mockup fidelity target: the concept's transfer-queue mockup (real BEM classes, lucide icons); reuse its state→color mapping. Add `data-testid`s and regenerate `tests/system/testid-catalog.md`.

## Acceptance criteria

- Panel shows all six states from the mockup, minimizes/restores, Clear Completed / Cancel All work.
- Reflects live backend events; matches the mockup layout.

## Test plan (TDD)

- **Unit (Vitest):** row rendering per state, store slice reducers, minimized toggle.
- **Integration (Python bridge):** drive a transfer and assert the panel updates.
- **Manual:** visual review against the mockup in light/dark. Run `pnpm build` for the tsc gate.

Issue #1559, opened 2026-07-17
fix(agent): 'Run Tests' red on develop — self_update_integration::applied_update_does_not_re_exec_on_the_next_idle fails on all platforms


armaxri:
`develop` is currently **red**: the `Run Tests` CI job fails on ubuntu, macOS and windows. This blocks every open PR, since branches inherit the failure through no fault of their own (first hit on #1558).

## The failure

```
test applied_update_does_not_re_exec_on_the_next_idle ... FAILED

thread 'applied_update_does_not_re_exec_on_the_next_idle' panicked at
agent/tests/self_update_integration.rs:493:5:
session did not close in time

test result: FAILED. 5 passed; 1 failed
```

Line 493 is the second of two 15s waits driving a session through the idle transition:

```rust
client.close_session(&session_id);
assert!(
    wait_until(Duration::from_secs(15), || client.session_count() == 0),
    "session did not close in time"
);
```

## Evidence it is not caused by any one PR

- **On `develop` itself**: run [29566868410](https://github.com/armaxri/termiHub/actions/runs/29566868410) (Code Quality, at the merge of #1556) — same test, same assertion.
- **On PR #1558** (a frontend-only file-browser routing change that cannot touch agent self-update): [29567190429](https://github.com/armaxri/termiHub/actions/runs/29567190429) — fails identically on all three platforms.

## Origin

The test was added by #1555 (`fix(agent): clear already-applied pending update at startup`, commit `9b1f80fc`, Closes #1551). The preceding `develop` run at #1548 was green, and #1554/#1556 are unrelated (autoformat hook / test-bridge verb), so #1555 is the prime suspect.

Note it is the *tail* of the test that fails: the binary swap and the agent's return after self-apply both succeed, so the self-update path itself works — it is the post-apply session close that does not settle within 15s.

## Scope

- [ ] Determine whether this is a real bug (session teardown genuinely not completing after a self-apply re-exec) or an under-tolerance timing assumption in the test.
- [ ] Fix accordingly — do **not** simply raise the timeout without establishing which of the two it is.
- [ ] Get `develop` back to green.

## Done when

`cargo test --workspace --all-features` passes on `develop` on all three platforms, and PRs stop inheriting a red `Run Tests`.

armaxri:
**Update: this is intermittent, not deterministic.**

[...]

PR #1565, opened 2026-07-17
test(system): Python-bridge integration test for the Transfer Queue panel (Closes #1532)

### LINKED ISSUE #1532: test(system): Python-bridge integration test for the Transfer Queue panel
Follow-up to #1337 (PR #1530), part of Epic #1331.

#1337 shipped the Transfer Queue panel with full Vitest unit coverage (mapper, store slice, row/controls per state, panel + minimized indicator) plus documented manual steps, but not the integration test named in the issue's test plan.

## Scope

- [ ] Add a Python-bridge system test (`tests/system/`) that drives a real transfer (SFTP or the FTP fixture) and asserts the panel updates: a `transfer-row` appears with rising percent, the state-appropriate controls render, Cancel/Clear Completed/Cancel All work, and Minimize collapses to `transfer-queue-indicator` which restores the panel.
- [ ] Use the `data-testid`s already in `tests/system/testid-catalog.md` (`transfer-queue`, `transfer-row`, `transfer-pause`/`resume`/`cancel`/`retry`/`remove`, `transfer-clear-completed`, `transfer-cancel-all`, `transfer-minimize`, `transfer-queue-indicator`).

## Done when

A `tests/system/` test drives a transfer end-to-end and asserts the Transfer Queue panel + minimized indicator behavior, passing via `./scripts/test-system-py.sh`.

armaxri:
## Summary

Adds the Python-bridge system test named in #1337's test plan (follow-up to PR #1530, part of Epic #1331). `tests/system/tests/test_transfer_queue.py` covers the Transfer Queue panel and its minimized indicator, in two suites split by what each can assert *deterministically*.

### `TestTransferQueueLiveTransfer` — a real SFTP transfer

Drives a **real transfer** over the `ssh-password` container and asserts the panel populates from it, exercising the whole chain inside the real app: backend `transfer-progress` emit → `listen` subscription → `useTransferEvents` → `applyTransferProgressToQueue` → render.

The entry point is the file browser's **copy → paste**. An SFTP→SFTP copy is implemented as "download to a local temp file, re-upload to the destination" (`useFileSystem.pasteEntry`), so one paste drives two real `sftp_download`/`sftp_upload` transfers.

**Why copy/paste and not Download/Upload:** it is the only transfer entry point the bridge can drive. Every other one (file-browser Download/Upload, the editor's "Download a copy") opens a native OS file dialog first, which the in-webview bridge cannot reach — the constraint already documented in `test_sftp_infra.py` and handled as guided-manual in `test_native_dialogs.py`.

Asserts: both rows reach `completed` at 100%, the row renders with Remove (and no pause/cancel/retry), the summary reads `2 done`, the remote **`path` from #1531 (PR #1543)** lands on both rows (source path on the download, destination path on the upload), and the pasted file really lands at full size — bytes moved, not just events emitted.

### `TestTransferQueuePanel` — deterministic per-state coverage

Drives the panel via `emitEvent` (#1545, PR #1556), injecting `transfer-progress` on the **real Tauri bus** — the app's own `listen` subscription and store fold run untouched; only the byte-moving backend is substituted. This buys determinism for the states a localhost transfer cannot be held in long enough to assert: rising percent (10% → 55% → 90%), `paused` ↔ `active`, `failed` with its `failed (2/3)` attempt counter, the exact control set per state, Cancel, Cancel All's enable/disable, Clear Completed keeping unfinished rows, and Minimize → `transfer-queue-indicator` → restore.

Needs no container, so it runs anywhere the app is built.

## Also

- **`ssh_password_fixtures`** (new, `conftest.py`): brings up the password-auth container *only*. `ssh_fixtures` also builds `ssh-keys`, which this suite never uses.
- The suite reads the browser's remote path from the store rather than `FilesUi.file_browser_path()` — see the follow-up note below.

## Verification

- `./tests/system/pytest.sh tests/test_transfer_queue.py` → **9 passed** against a real `pnpm tauri build --debug`, with the `ssh-password` container live.
- Machinery lane (`-m "not integration"`, what PR CI runs) → 193 passed; these 9 are correctly deselected as `integration`.
- `npx tsc --noEmit` clean.
- **Confirmed the tests fail without the behaviour they cover** — each sabotage was rebuilt and run, then reverted:
  - reverting #1543's mapper line (`path: progress.path ?? prev?.path` → `prev?.path`) turns the live suite red: `assert None == '/home/testuser/payload-….bin'`.
  - gating off the `paused` branch of `TransferControls` turns the panel suite red: `assert False = exists('transfer-resume')`.

## Note for reviewers

These suites are `integration`-marked, so **PR CI does not run them** — the machinery lane only checks that they collect. They were verified locally as above.

## Done when

A `tests/system/` test drives a transfer end-to-end and asserts the Transfer Queue panel + minimized indicator behavior, passing via the harness. ✅

Closes #1532

### DIFF
--- a/tests/system/conftest.py
+++ b/tests/system/conftest.py
@@ -223,6 +223,17 @@ def ssh_fixtures():
 
 
 @pytest.fixture(scope="session")
+def ssh_password_fixtures():
+    """Password-auth SSH container only (port 2201).
+
+    ``ssh_fixtures`` also brings up ``ssh-keys``; suites that never authenticate
+    with a key should ask for this instead, so they neither wait for that image
+    to build nor skip when only *it* fails to.
+    """
+    return _ensure_ssh_services([(SSH_PASSWORD_SERVICE, SSH_PASSWORD_PORT)])
+
+
+@pytest.fixture(scope="session")
 def ssh_x11_fixtures():
     """X11-forwarding SSH container (port 2208).
 
--- a/tests/system/tests/test_transfer_queue.py
+++ b/tests/system/tests/test_transfer_queue.py
@@ -1,0 +1,494 @@
+"""Transfer Queue panel integration tests (issue #1532).
+
+The Vitest suite from #1337 (PR #1530) covers the mapper, the store slice and
+every row/control permutation in isolation. What it cannot cover is the wiring:
+that a **real backend transfer** reaches the panel at all — the
+``transfer-progress`` event → ``onTransferProgress`` ``listen`` subscription →
+``useTransferEvents`` → ``applyTransferProgressToQueue`` → ``TransferQueue``
+render chain, running inside the real app. That is what this suite adds.
+
+Two suites, split by what each can assert *deterministically*:
+
+``TestTransferQueueLiveTransfer``
+    Drives a **real SFTP transfer** over the Docker ``ssh-password`` fixture and
+    asserts the panel populates from it. The entry point is the file browser's
+    **copy → paste**: an SFTP→SFTP copy is implemented as "download to a local
+    temp file, re-upload to the destination" (``useFileSystem.pasteEntry``), so a
+    single paste drives two real ``sftp_download`` / ``sftp_upload`` transfers
+    and emits real ``transfer-progress`` events — including the remote ``path``
+    added by #1531 (PR #1543), which this suite asserts reaches the row.
+
+    Copy/paste is used deliberately: it is the **only** transfer entry point the
+    bridge can drive. Every other one (file-browser Download/Upload, the
+    editor's "Download a copy") opens a native OS file dialog first, which the
+    in-webview bridge cannot reach — the same constraint documented in
+    ``test_sftp_infra.py`` and handled as guided-manual in ``test_native_dialogs.py``.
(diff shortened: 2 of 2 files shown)


armaxri:
Follow-ups filed from this work:

- **#1568** — `FilesUi`/`SftpUi` still target the removed `file-browser-current-path` testid, so `connect_sftp_browser()`, `enter_directory()`, `file_browser_path()` and friends all time out. This suite works around it by reading `currentPath` from the store; the workaround should go once the helpers are fixed.
- **#1573** — the Transfer Queue row for a paste-upload is named after the internal temp file (`termihub-paste-<ts>-<name>`), because `sftp_upload` derives `file_name` from the *local* path. This suite asserts the current behaviour, so that assertion needs updating when #1573 lands.

Note on CI: `Run Tests (ubuntu-latest)` failed on the first push with `self_update_integration::applied_update_does_not_re_exec_on_the_next_idle` — that is #1559, already red on `develop@bbe9797e` (this branch's base) with an identical panic, and unrelated to this Python-only diff. It passed on the re-run, so it appears intermittent.

Issue #1569, opened 2026-07-17
ci: integration-marked system tests never execute in CI (-m "not integration")


armaxri:
## What

The System-Test Harness job runs:

```yaml
# .github/workflows/code-quality.yml:221
run: ./tests/system/pytest.sh -m "not integration" -q
```

So every test marked `integration` is excluded from CI. Nothing else runs them. They are merged unexecuted.

## Why it matters now

Two live Python-bridge E2Es landed today into that lane and have **never been run anywhere**:

- `tests/system/tests/test_agent_update_banner_live.py` (#1520, PR #1564)
- the Transfer Queue panel test (#1532, in flight)

Both were written carefully — the #1520 author verified each assertion against source and caught a real bug that way (a dismiss test asserted sticky dismissal, but `setAgentUpdateAvailable` deliberately clears the dismissed flag). But source-reading is not execution. An unrun test suite provides confidence without evidence, which is worse than no suite.

## Options

1. Run the integration lane in CI (needs a container engine on the runner).
2. Run it on a schedule / nightly rather than per-PR, if runtime is the concern.
3. Keep excluding it, but make that explicit and gate merges of integration tests on a documented local run.

Whichever way, the current state — tests that exist, are marked, and execute nowhere — should not be the resting state.

## Note on local execution

Local runs are possible: both `docker` and `podman` are available on the dev machine. An agent concluded "Docker unavailable" while Docker happened to be stopped and did not check podman, so verify the engine before assuming the lane is unrunnable locally.

Filed by the dev coordinator from #1520 / PR #1564.

armaxri:
## Concrete evidence of the cost, found the same day

#1568 is exactly what this issue predicts, and it was found by accident rather than by CI.

`FilesUi`/`SftpUi` target `file-browser-current-path`, a testid that **no longer exists in `src/`** — the path bar became the breadcrumb `FileBrowserPathBar`. Verified independently by the coordinator: zero hits in `src/`, but still referenced from the harness:

```
tests/system/tests/test_transfer_queue.py
tests/system/tests/test_ssh_password_extended.py
tests/system/tests/test_file_browser_local.py
```

That silently breaks `connect_sftp_browser()`, `enter_directory()`, `file_browser_path()`, `navigate_up()` and `open_file_browser()` — meaning `test_sftp_infra.py` and friends **cannot pass as written**.

[...]

Issue #1528, opened 2026-07-14
ci/tooling: stop the testid catalog from going stale and breaking unrelated PRs' CI


armaxri:
The committed `tests/system/testid-catalog.md` + its CI `--check` gate (Frontend Code Quality + System-Test Harness) has repeatedly broken CI on **unrelated** PRs (#1492, #1525, and others). Two independent failure modes:

1. **Develop-advance staleness (structural):** the catalog is a single global generated file. Whenever any PR merges a testid change, *every* open branch's committed catalog is now stale, so its `--check` fails until it merges develop — even for PRs that touch no testids. This is the same churn the repo already eliminated for `CHANGELOG.md` via per-branch change fragments.
2. **Autoformat-hook trimming:** the PostToolUse autoformat/incremental path has produced *trimmed* catalogs (dropping forwarded-prop entries like `server-item-*`), which agents then commit, failing the forwarded-prop catalog test.

Each incident needs a manual "merge develop + `python scripts/build-testid-catalog.py` + push" fix.

## Options (pick one in the PR)
- [ ] **Preferred: generate-and-verify in CI instead of diffing a committed file.** Have the CI job run `build-testid-catalog.py` against the merged tree and validate coverage/consistency there; stop committing the catalog (or keep it as a non-gating, regenerated artifact). Removes the staleness gate entirely.
- [ ] **Or: make the autoformat hook always run the FULL `build-testid-catalog.py`** (never an incremental trim) so a saved catalog is always complete — fixes mode 2, not mode 1.
- [ ] **Or: treat the catalog like `CHANGELOG.md`** — don't commit the aggregate; regenerate at need.

## Done when
A PR that changes no testids never fails CI due to catalog staleness, and the autoformat hook can't commit a trimmed catalog. No more per-PR "regenerate catalog" fixups.

armaxri:
Related pre-existing issue: #1526, which is the concrete **generator-divergence** cause (autoformat hook's `regen-testid-catalog.mjs` vs the `build-testid-catalog.py` CI gate). That covers the 'trimming' failure mode here; this ticket additionally covers the **develop-advance staleness** mode (a merged testid change staleing every open branch). Whichever fix lands, the two should be reconciled — #1526's 'single source of truth' plus this ticket's 'generate-and-verify in CI / stop committing the catalog' are complementary.

PR #1575, opened 2026-07-17
test(sftp): point the file-browser path helpers at the real path bar (Closes #1568)

### LINKED ISSUE #1568: test(system): FilesUi/SftpUi target the removed file-browser-current-path testid
Found while implementing #1532 (PR #1565).

`FilesUi.CURRENT_PATH` is `file-browser-current-path`, which **no longer exists anywhere in `src/`** — the path bar became the breadcrumb `src/components/Sidebar/FileBrowserPathBar.tsx` (`file-browser-crumb`, `file-browser-path-edit`, `file-browser-path-input`).

Every helper built on it is dead: `file_browser_path()` (always `""`), `open_file_browser()`, `wait_for_path_contains()`, `enter_directory()`, `navigate_up()`, and `SftpUi.connect_sftp_browser()` all time out.

Confirmed live: with the `ssh-password` container healthy and the store showing `sftpStatus: connected`, `currentPath: /home/testuser` and populated `fileEntries`, the bridge still reports `no element with data-testid="file-browser-current-path"`.

**Why unnoticed:** these suites are `integration`-marked and PR CI runs `-m "not integration"`, so nothing exercises them per-PR. Related in spirit to #1526 / #1528.

**Impact:** `tests/system/tests/test_sftp_infra.py` and friends cannot pass as written.

**Fix:** point the helpers at the breadcrumb path bar, or read `currentPath` from the store (what the bar renders from). #1565 works around this locally by reading the store; that workaround should go once the helpers are fixed.

Worth considering a lane that actually runs the container-backed suites, or the next refactor rots them the same way.

armaxri:
The `FilesUi` / `SftpUi` helpers target a `file-browser-current-path` testid that no longer exists anywhere in `src/`. The path bar became the breadcrumb `FileBrowserPathBar` and the id went with the old markup.

It fails **silently**: `exists()` returns False and `get_text()` returns `""`, so `connect_sftp_browser()`, `open_file_browser()`, `file_browser_path()`, `enter_directory()` and `navigate_up()` simply time out rather than erroring usefully. PR CI only ever runs `-m "not integration"`, so integration tests are checked to *collect*, never to *run* — the rot was invisible. That dark lane is #1569 and is out of scope here.

## The fix

**Restore the id on the breadcrumb `<nav>`, read its `title`.** The nav already carries `title={currentPath}` — the full path verbatim, and what the user sees on hover. `get_text()` is not usable here: the crumbs render the path as separate labelled buttons, so it would return the labels run together rather than a path.

**Read the bar, not the store.** The alternative was the store's `currentPath`. That would be wrong: `useFileBrowser` selects `currentPath` from one of three sub-hooks (local / sftp / session) by mode, and the store field is the **SFTP** one only — a store read returns the wrong path for a local browser, silently. The bar renders whichever mode is active, so reading it is correct for all three.

**One implementation, not two.** The read was copy-pasted into both mixins, which is precisely why a single markup change rotted both at once. It now lives in a shared `FileBrowserPathReads` that `FilesUi` and `SftpUi` inherit — the two suites that mix both get one implementation instead of two that can drift.

Also folds in two further copies of the same rot:
- `test_ssh_password_extended.py` hand-rolled the same flow inline with the dead literal; it now mixes `SftpUi` and uses the shared read (kept explicit rather than calling `connect_sftp_browser()`, which would swallow the prompt assertion the test exists to make).
- `test_transfer_queue.py` (#1532, PR #1565) worked around the broken helper by reading the store. With the helper fixed the workaround is unnecessary and is removed — its 9 tests still pass.

The testid catalog is generated and gitignored (#1528); regenerating picks the id up under `FileBrowserPathBar.tsx`.

## Test plan

CI cannot validate this fix (#1569), so it was run locally against a real build with the Docker SSH fixtures live:

- `test_sftp_infra.py` — the bar for this work
- `test_transfer_queue.py` — confirms the removed workaround
- `test_file_browser_local.py` — the local browser path
- `test_ssh_password_extended.py`, `test_ssh_sftp_cwd.py`
- Frontend unit: 226 passed / 33 files
- Python non-integration: 193 passed

Results are reported in the PR thread.

Closes #1568

### DIFF
--- a/src/components/Sidebar/FileBrowserPathBar.tsx
+++ b/src/components/Sidebar/FileBrowserPathBar.tsx
@@ -74,7 +74,12 @@ export function FileBrowserPathBar({ currentPath, onNavigate }: FileBrowserPathB
 
   return (
     <div className="file-browser__path-bar">
-      <nav className="file-browser__breadcrumbs" aria-label="Path" title={currentPath}>
+      <nav
+        className="file-browser__breadcrumbs"
+        aria-label="Path"
+        title={currentPath}
+        data-testid="file-browser-current-path"
+      >
         {crumbs.map((crumb, i) => (
           <span className="file-browser__crumb-group" key={crumb.path}>
             {i > 0 && <ChevronRight size={11} className="file-browser__crumb-sep" aria-hidden />}
--- a/tests/system/termihub_harness/ui/files.py
+++ b/tests/system/termihub_harness/ui/files.py
@@ -7,8 +7,9 @@ and create files/folders via the inline toolbar inputs. It builds on
 both alongside :class:`~termihub_harness.SystemTest`.
 
 This is the local counterpart to :class:`~termihub_harness.ui.SftpUi` (remote SFTP
-browsing): both read ``file-browser-current-path`` and ``file-row-<name>``, but a
-local browser needs no password prompt and follows the terminal's CWD.
+browsing): both read ``file-browser-current-path`` and ``file-row-<name>`` — via
+the shared :class:`FileBrowserPathReads` — but a local browser needs no password
+prompt and follows the terminal's CWD.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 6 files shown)


armaxri:
## Local run results

CI cannot validate this fix (#1569 — PR CI only runs `-m "not integration"`), so it was run locally on macOS against a real debug build with the Docker SSH fixtures live. Verified the tests actually **ran** rather than silently skipping (the compose fixture `pytest.skip`s when no runtime is available, so "no failures" is not the same as "passed").

| Suite | Result |
| --- | --- |
| `test_sftp_infra.py` | **7 passed** — the bar for this issue |
| `test_transfer_queue.py` | **9 passed** — confirms removing the #1532 store workaround |
| `test_ssh_sftp_cwd.py` | **4 passed** |
| `test_ssh_password_extended.py` | **2 passed** |
| `test_sftp_close_all_on_quit.py` + `test_ssh_monitoring_settings.py` + `test_external_app.py` | **6 passed**, 7 skipped (guided-manual, need `--manual`) |
| `test_file_browser_local.py` | **10 passed / 6 failed** → see #1578 below |
| Frontend unit | 226 passed (33 files) |
| Python non-integration | 193 passed |

`test_ssh_sftp_cwd` is the most direct evidence: it asserts the displayed path follows `cd`, which exercises `file_browser_path()` / `wait_for_path_contains()` end to end. None of these could pass before — the helper returned `""` forever.

### Evidence for reading the bar rather than the store

The failure artifact from a local-browser test captures exactly the trap the store approach would have fallen into:

```
fileBrowserMode  = local
localCurrentPath = /Users/arne     ← what the browser actually shows
currentPath      = /              ← the store field #1532 read
```

A store read returns `/` for a local browser while the bar shows `/Users/arne` — wrong, and silently so. Reading the bar is correct in all three modes.

### The 6 remaining `test_file_browser_local` failures are not this bug — filed as #1578

All 6 are `wait_for_file_row` timeouts with a distinct root cause. The entry is **not** missing: the artifact's `state.json` shows `localFileEntries` holding 104 entries **including** the created one, and the terminal log confirms the `mkdir` ran. But the row is not in the DOM — the list is virtualized (`FileBrowser.tsx:937`: *"Only the visible window of rows (+overscan) is mounted"*), and the suite browses the real `$HOME`, where the entry sorts below the fold.

Container-based suites are unaffected because their home has few entries, so every row fits the window. That is a separate defect in the row helpers (not the path helpers), so it is #1578 rather than scope creep here.

### CI note

`Run Tests` first came back red on all three platforms with `armed_hook_makes_the_deferred_busy_path_reachable` (`agent/tests/deferred_update_hook_integration.rs`). This diff contains **zero** Rust and does not touch `agent/`. It is a flake in the same timing-sensitive agent update-hook area as #1559 (develop itself is currently red on `applied_update_does_not_re_exec_on_the_next_idle`). Re-running the unchanged diff went green.

armaxri:
## Local verification (this lane is dark in CI)

PR CI runs the system harness with `-m "not integration"` (`.github/workflows/code-quality.yml:221`), so the suites this PR fixes are only ever **collected**, never run (#1569). Verified locally instead — macOS, debug build rebuilt from this branch (so the new testid is actually in the bundle), live Docker fixtures, `dev.local.json` → `termihub-test-3`:

| Suite | Result |
| --- | --- |
| `test_sftp_infra.py` | **7 passed** |
| `test_transfer_queue.py`, `test_ssh_password_extended.py`, `test_ssh_sftp_cwd.py`, `test_sftp_close_all_on_quit.py` | passing (26 passed overall in the combined run) |
| `test_file_browser_local.py` | 6 failed — **pre-existing**, see below |

### The 6 `test_file_browser_local.py` failures are not from this PR

On `origin/develop` the testid does not exist at all, so `open_file_browser()` times out at the **first** step and those tests never reach their row assertions. This PR fixes the path read — proved by `test_shows_an_absolute_current_path`, `test_toolbar_is_visible_for_a_local_terminal` and `test_up_button_navigates_to_the_parent` now passing — which **unmasks** a separate defect: entries created through the terminal never list. Filed as #1582 with the diagnosis (not a stale testid, not a missing refresh — likely a cwd mismatch in `ShellFsUi.touch_home`'s relative `touch`).

### No further testid rot

Swept every testid-like literal in `tests/system/` against a freshly generated `scripts/build-testid-catalog.py` catalog (619 literal + 164 dynamic ids): **zero unmatched**. #1568 was the only stale one.

### CI note

`agent/tests/deferred_update_hook_integration.rs::armed_hook_notifies_the_client_on_attach` failed once on ubuntu with `Os { code: 26, kind: ExecutableFileBusy, message: "Text file busy" }` — a spawn race in the agent test harness, unrelated to this PR (which touches zero Rust). macOS/Windows were fail-fast **cancellations**, not real failures. Re-run.

PR #1584, opened 2026-07-17
fix(testing): report an explicit per-suite verdict in test.sh (Closes #1572)

### LINKED ISSUE #1572: fix(testing): scripts/test.sh can report a passing suite and ELIFECYCLE at the same time
## What

`./scripts/test.sh` was observed reporting **both** success and failure in one run:

```
3223 passed
...
ELIFECYCLE
```

A run that says "3223 passed" and then exits non-zero is ambiguous to a human and, more importantly, to an agent deciding whether its change is green.

## How it surfaced

During #1335 (PR #1558): a new SSH regression test caused an unhandled error — entering `sftp` mode fires a real auto-connect, and the mocked `sftp_list_dir` returned `undefined`. The underlying test problem was found and fixed before merge. What remains is the **reporting** behaviour: the suite printed a full pass line alongside the lifecycle failure.

## Why it matters

The pass line is the one a reader anchors on. Anything that makes "did my suite pass?" require interpreting two contradictory signals will eventually be read wrong — the failure is easy to miss under a large green summary.

Worth deciding whether the unhandled-error path should suppress the pass summary, or the summary should carry the non-zero exit explicitly.

Filed by the dev coordinator from #1335 / PR #1558.

armaxri:
## The decision

The issue offers two options. Only one of them is `test.sh`'s to take.

**Rejected — "suppress the pass summary".** That line is vitest's, not the script's. Suppressing it would mean owning a custom reporter, and it would be wrong regardless: the 3223 tests genuinely *did* pass, and vitest already prints `Errors  1 error` in that same summary block. Deleting accurate output to fix a reading problem is the wrong trade. **The fix is to answer the pass line, not to hide it.**

**Taken — "carry the non-zero exit explicitly".** This is squarely `test.sh`'s job, and it is where the defect actually is.

## The defect is sharper than the issue states

`test.sh` already got the *exit code* right (`FAILED=1`, `exit 1`) — that part was never broken and is now covered by tests so it stays that way. The real problem is the **verdict**: success printed `PASS`, failure printed **nothing at all**. So a failing suite's last word was the runner's own contradicting summary, and then `cargo test` ran for minutes and scrolled it away. The final `SOME TESTS FAILED.` never said *which* suite.

Reproduced with stubbed runners on the parent commit:

```
=== Frontend: Vitest ===
      Tests  3223 passed (3223)
 ELIFECYCLE  Test failed. See above for more details.
                                        <- frontend failure ends here, unmarked
=== Rust workspace: cargo test ===
test result: ok
PASS                                    <- the only verdict on screen
SOME TESTS FAILED.                      <- which one?
```

A reader scrolling back to find the failure lands on `3223 passed` beside the frontend section and reasonably concludes the frontend was fine and Rust broke. That is the mixed signal.

## After

```
=== Frontend: Vitest ===
      Tests  3223 passed (3223)
 ELIFECYCLE  Test failed. See above for more details.
FAIL: Frontend: Vitest (exit 1)

=== Rust workspace: cargo test ===
PASS: Rust workspace: cargo test

SOME TESTS FAILED:
  - Frontend: Vitest (exit 1)
```

The count describes the tests; the exit code describes the run. Only the exit code decides, and the verdict now sits immediately after it so the two cannot be read apart.

## Notes

- **Phrasing follows existing precedent** — `test-system-{linux,mac,windows}.sh` already print `SOME TESTS FAILED (exit code: $TEST_EXIT)`. This aligns `test.sh` with its own siblings rather than inventing a style.
- **`set -e` safety**: each suite runs under `|| status=$?` so a red suite records a verdict instead of aborting the script — one failing suite must not hide the other's result. Covered by a test.
- **`test.cmd`** is kept in step with the same verdict format.
- **No change fragment**: `scripts/test.sh` is a dev helper, not user-facing app behaviour.

## Verification

TDD — the five bug-describing tests fail on the parent commit and pass after the fix (3 baseline tests pass on both, pinning the already-correct exit-code behaviour). Tests live in `tests/system/tests/test_test_script.py`, following the `test_runner_script.py` precedent, and run in the **machinery** CI lane (no app build, no Docker). `pnpm`/`cargo` are stubbed on `PATH` and the stub reproduces the real contradicting-summary shape, so the guard covers the actual bug rather than a simplification of it.

Also ran the real `./scripts/test.sh` end-to-end: frontend `3270 passed` -> `PASS: Frontend: Vitest`. The run also surfaced an unrelated pre-existing failure, which the new output attributed correctly on the first read — `FAIL: Rust workspace: cargo test (exit 101)`. Filed as a follow-up; this branch changes no Rust (see below).

## Follow-ups filed

- **#1585** — `docker_spawn` integration test fails reproducibly against Docker Desktop on macOS (Docker up, `alpine` present, 3/3 runs). Unrelated to this branch, which changes no Rust. Found by running the real `./scripts/test.sh` while verifying this fix — and it was this PR's new output that named the culprit suite on the first read.

### DIFF
--- a/scripts/test.cmd
+++ b/scripts/test.cmd
@@ -13,19 +13,37 @@ if not exist node_modules (
 
 set FAILED=0
 
+REM A runner's own summary is not the verdict. Vitest exits non-zero when the run
+REM hit an unhandled error even though every test passed, so its summary can read
+REM "3223 passed" on a run that failed (#1572): the count describes the tests, the
+REM exit code describes the run. Only the exit code decides here, and the verdict
+REM is printed right after it so the two cannot be read apart. Mirrors test.sh.
 echo === Frontend: Vitest ===
 call pnpm test
-if errorlevel 1 set FAILED=1
+set FRONTEND_STATUS=%errorlevel%
+call :verdict %FRONTEND_STATUS% "Frontend: Vitest"
 
 echo.
 echo === Rust workspace: cargo test ===
 cargo test --workspace --all-features
-if errorlevel 1 set FAILED=1
+set RUST_STATUS=%errorlevel%
+call :verdict %RUST_STATUS% "Rust workspace: cargo test"
 
 echo.
 if %FAILED%==1 (
-    echo SOME TESTS FAILED.
+    echo SOME TESTS FAILED:
+    if not "%FRONTEND_STATUS%"=="0" echo   - Frontend: Vitest ^(exit %FRONTEND_STATUS%^)
+    if not "%RUST_STATUS%"=="0" echo   - Rust workspace: cargo test ^(exit %RUST_STATUS%^)
     exit /b 1
-) else (
-    echo ALL TESTS PASSED.
 )
+echo ALL TESTS PASSED.
+exit /b 0
+
+:verdict
+if "%~1"=="0" (
+    echo PASS: %~2
+) else (
+    echo FAIL: %~2 ^(exit %~1^)
+    set FAILED=1
+)
+goto :eof
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -17,25 +17,45 @@ if [ ! -d node_modules ]; then
 fi
 
 FAILED=0
+FAILED_SUITES=""
 
+# Print the verdict for a suite that has just run, and remember any failure.
+#
+# A runner's own summary is not the verdict. vitest exits non-zero when the run
+# hit an unhandled error even though every test passed, so its summary can read
+# "3223 passed" on a run that failed (#1572): the count describes the tests, the
+# exit code describes the run. Only the exit code decides here, and the verdict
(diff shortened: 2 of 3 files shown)


armaxri:
Merged with one red lane: `Run Tests` failed on `self_update_integration::applied_update_does_not_re_exec_on_the_next_idle`, tracked as issue 1559 and red on `develop` itself right now (merge commits `b98ee55a`, `7b6aa222`, `77b1a784` all failed the same test). Re-ran once; identical result.

Not attributable to this branch: the diff is `scripts/test.sh`, `scripts/test.cmd` and one Python test — zero Rust — so it cannot affect `agent/tests/self_update_integration.rs`. Every lane that reflects this change is green, including **System-Test Harness (machinery)**, where the new tests run.

The macOS and Windows `Run Tests` lanes show as failed but were **cancelled** by matrix fail-fast after ubuntu tripped, not independently red.

Issue #1591, opened 2026-07-17
test(system): editor status-bar items never clear when switching to a terminal tab


armaxri:
Follow-up to #1582 (PR pending). Found while verifying the #1582 fix — **pre-existing on `develop`, not caused by that work**.

## What

Two `test_editor.py` cases fail locally under `-m integration`:

```
FAILED tests/test_editor.py::TestEditor::test_status_bar_items_hide_on_a_terminal_tab
FAILED tests/test_editor.py::TestEditor::test_status_bar_items_return_when_switching_back_to_the_editor
```

Both fail the same way:

```
AssertionError: timed out waiting for the editor status to clear on the terminal tab (last error: None)
termihub_harness/systemtest.py:150
```

## Why this is not #1582

Verified by running both against `f4ae55ce` (develop's `ui/files.py`, i.e. the #1582 fix reverted): **they fail identically**. They also get well past `wait_for_file_row` — the editor opens fine and the status bar populates; the failure is that `editorStatus` never clears once a terminal tab becomes active.

## Why it matters

These were not in the known-failing set. The #1582 issue reported the six `test_file_browser_local.py` cases as the only remaining failures; these two are additional, and reproduce on a clean `develop`.

## Scope

- [ ] Determine whether `editorStatus` genuinely fails to clear on terminal-tab activation (an app bug) or the test's expectation/wait is wrong (a test bug).
- [ ] Fix the root cause accordingly.
- [ ] Confirm both cases pass locally: `./scripts/test-system-py.sh --debug -m integration -k editor`

## Done when

`-m integration -k editor` is green locally.

## Note

This lane is dark in PR CI (`-m "not integration"`, #1569) — it must be verified by a local run.

armaxri:
Fixed by PR #1614 (merged into develop). The two status-bar cases were a test bug: the tests used ensure_terminal() to 'switch to a terminal tab', but it does not change the active tab when a terminal already exists, so editorStatus (correctly) never cleared. They now switch to the terminal tab explicitly. No app change needed.

PR #1592, opened 2026-07-17
fix(ui): filter the file browser to find rows outside the virtual window (Closes #1582)

### LINKED ISSUE #1582: test(system): test_file_browser_local entries created via the terminal never list
Follow-up to #1568 (PR #1575).

## What

With the `file-browser-current-path` testid rot fixed in #1568, `test_file_browser_local.py` now gets **past** `open_file_browser()` and reaches its real assertions. Six of them then fail — a defect that was previously **masked** by the earlier failure, not a regression from #1575.

Local run on this branch (macOS, debug build, live `ssh-password` fixture):

```
6 failed, 26 passed, 525 deselected in 225.88s
FAILED test_file_browser_local.py::TestFileBrowserLocal::test_lists_entries_from_the_current_directory
FAILED test_file_browser_local.py::TestFileBrowserLocal::test_double_click_enters_a_subdirectory
FAILED test_file_browser_local.py::TestFileBrowserLocal::test_create_file_via_inline_input
FAILED test_file_browser_local.py::TestFileBrowserLocal::test_create_folder_via_toolbar
FAILED test_file_browser_local.py::TestFileBrowserLocal::test_file_context_menu_offers_edit_rename_delete
FAILED test_file_browser_local.py::TestFileBrowserLocal::test_context_menu_delete_removes_the_file
```

All six fail the same way:

```
E  AssertionError: timed out waiting for file row 'e2e_fb_sys-entry-5e010bec.txt'
   termihub_harness/ui/files.py:125
```

Reproduces in isolation:

```bash
./scripts/test-system-py.sh --debug --skip-build -m integration -k "file_browser_local and lists_entries" -x
```

## Why it is pre-existing, not caused by #1575

On `origin/develop`, `FileBrowserPathBar.tsx` renders **no** `file-browser-current-path` testid, and `FilesUi.open_file_browser()` waits on `get_text(CURRENT_PATH)`. Every test calling `_fresh_home_browser()` therefore times out at the **first** step on develop and never reaches the row assertions. #1575 fixes the path read (proved by `test_shows_an_absolute_current_path`, `test_toolbar_is_visible_for_a_local_terminal` and `test_up_button_navigates_to_the_parent` now passing), which is what exposes this.

## The common factor

Every failing case waits for a **newly created** entry to appear; every passing case does not. Ruled out already:

- **Not a stale testid.** A sweep of all testid-like literals in `tests/system/` against a freshly generated `scripts/build-testid-catalog.py` catalog (619 literal + 164 dynamic) returns **zero** unmatched ids. `file-row-*` is present and correct.
- **Not a missing refresh.** `wait_for_file_row` already clicks `file-browser-refresh` every 2s while polling (`refresh_every: float = 2.0`).

So the entry genuinely is not in the directory the browser lists. Prime suspect: `ShellFsUi.touch_home()` runs a **relative** `touch <name>` in the shell's cwd (`tests/system/termihub_harness/ui/shell_fs.py:49`), while the browser lists whatever path it resolved — if the shell's cwd and the browser's directory disagree after `restart_app()` + `ensure_terminal()`, the file lands somewhere the browser is not showing. Worth also confirming the `run_command` actually executes before the wait begins.

## Scope

- [ ] Determine whether the shell's cwd and the browser's listed directory actually agree after `_fresh_home_browser()` (log both).
- [ ] Fix the mismatch — either author fixtures at the browser's real path, or make the helpers resolve an absolute home path rather than relying on a relative `touch`.
- [ ] Confirm all of `test_file_browser_local.py` passes locally with `-m integration` against a real build.

## Done when

`./scripts/test-system-py.sh --debug -m integration -k file_browser_local` is fully green locally.

## Note

This lane is dark in PR CI (`-m "not integration"`, `.github/workflows/code-quality.yml:221`) — integration tests are only ever collected, never run (#1569). This must be verified by a local run.

armaxri:
## Root cause — a harness bug, not an app bug

The file-browser row list is **virtualized** (`@tanstack/react-virtual`): only the visible window of rows plus `ROW_OVERSCAN` is mounted — roughly the first ~33 at the default sidebar height (`ROW_HEIGHT = 24`, `ROW_OVERSCAN = 8`). `wait_for_file_row` assumed every listed entry had a DOM node. An entry sorting past that window has **no node at all**, so no amount of waiting or refreshing could ever reveal it.

Measured from the failing run's own `state.json`: the store held **104 entries** for `/Users/arne` and the sentinel sorted to **index 80**. The data layer was correct the whole time — `localCurrentPath` was right and `localFileEntries` *contained the file*. Only the DOM lacked the row.

That is exactly why the six failures were the six cases waiting on a **newly created** entry, while every case that only reads the path passed.

## The issue's prime suspect was wrong

#1582 suspected `ShellFsUi.touch_home`'s relative `touch` landing outside the browsed directory. It isn't:

- The captured `terminal.txt` shows the command already absolute and succeeding at `~`: `touch "$HOME/e2e_fb_sys-entry-5e010bec.txt"`.
- **Four of the six failures never touch the terminal at all** — `test_create_file_via_inline_input`, `test_create_folder_via_toolbar` and both context-menu cases create their entry through the browser's *own* New File/Folder input and fail identically. The creation route was never the variable; list position was.

## The fix

Type the entry name into the browser's **filter** box — the affordance a real user reaches for to find a file in a long listing — which narrows `displayEntries` until the row falls inside the mounted window. No sleeps, no bumped timeouts, no retry wrapper.

The filter survives a refresh (only navigation calls `resetNavState`), so it composes with the existing refresh polling, and it is deliberately **left applied** on return: the row's node exists only because of it, so clearing it would unmount the very row the caller is about to click. `clear_entry_filter()` is exposed for tests that need the full listing.

Fixed centrally in `wait_for_file_row`, so the SFTP, transfer-queue, editor and external-app suites that share it get the same robustness.

## Verification — local, against a forced rebuild

CI cannot validate this (`-m "not integration"`, #1569), so it was proven locally.

```
before:  6 failed, 10 passed   (-k file_browser_local, --rebuild)
after:  16 passed, 0 failed
```

No regressions in the other consumers of the shared wait:

```
-k "sftp_infra or transfer_queue" --fixtures ssh-password  →  16 passed
```

## Regression test

`tests/system/tests/test_files_ui.py` drives `FilesUi` against a fake driver modelling the virtual window + filter. It fails against the old wait with the production symptom (`timed out waiting for file row 'target.txt'`) and passes with the fix.

It sits in the **machinery lane**, so — unlike the integration lane (#1569) — it actually runs in CI.

## Follow-up filed

- #1591 — two `test_editor.py` status-bar cases fail on a clean `develop` (`editorStatus` never clears on a terminal tab). Verified pre-existing by re-running against `f4ae55ce` with this fix reverted; unrelated to this work.

Also seen once and not reproduced in two further runs: `test_connection_editor.py::test_new_connection_in_folder_via_context_menu` — noted as a possible flake, not filed.

## Test plan

- [x] `./scripts/test-system-py.sh --debug -m integration -k file_browser_local` → 16 passed
- [x] `./scripts/test-system-py.sh --debug --fixtures "ssh-password" -m integration -k "sftp_infra or transfer_queue"` → 16 passed
- [x] `cd tests/system && ./pytest.sh -k files_ui` → 6 passed (and red without the fix)
- [x] `./scripts/check.sh` → ALL CHECKS PASSED

### DIFF
--- a/tests/system/termihub_harness/ui/files.py
+++ b/tests/system/termihub_harness/ui/files.py
@@ -23,7 +23,12 @@ from .base import HarnessMixin
 
 
 def file_row_testid(name: str) -> str:
-    """The ``data-testid`` of the file-browser row for an entry ``name``."""
+    """The ``data-testid`` of the file-browser row for an entry ``name``.
+
+    Present in the DOM only while the row is inside the virtualizer's mounted
+    window — see :meth:`FilesUi.wait_for_file_row`, which filters to the entry
+    rather than assuming every listed entry has a node.
+    """
     return f"file-row-{name}"
 
 
@@ -70,6 +75,8 @@ class FilesUi(FileBrowserPathReads):
     # Stable file-browser toolbar / inline-input testids (entry-name-free).
     UP = "file-browser-up"
     REFRESH = "file-browser-refresh"
+    FILTER = "file-browser-filter"
+    FILTER_CLEAR = "file-browser-filter-clear"
     NEW_FILE = "file-browser-new-file"
     NEW_FILE_INPUT = "file-browser-new-file-input"
     NEW_FILE_CONFIRM = "file-browser-new-file-confirm"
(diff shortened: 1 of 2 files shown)


armaxri:
CI note: the only genuine failure is `self_update_integration::applied_update_does_not_re_exec_on_the_next_idle` on ubuntu — the known flake #1559, owned elsewhere. The macOS and Windows "Run Tests" jobs did not fail on their own; both were **canceled by matrix fail-fast** when ubuntu failed (`The operation was canceled` mid-step), which is why they show red.

Re-ran the failed jobs once; ubuntu reproduced the same #1559 panic at `agent/tests/self_update_integration.rs:493`.

This diff touches **two Python files only** (`tests/system/termihub_harness/ui/files.py`, `tests/system/tests/test_files_ui.py`), so it cannot affect a Rust agent test. `System-Test Harness (machinery)` — the lane that actually runs this PR's new regression test — is **green**.

Merging on that basis; the real verification for this change is the local integration run recorded in the PR body (6 red → 16 green).

PR #1598, opened 2026-07-17
feat(sftp): support the file editor in session-layer file browsing (Closes #1557)

### LINKED ISSUE #1557: feat(sftp): support the file editor in session-layer file browsing (FTP, Docker, agent sessions)

armaxri:
Follow-up to #1335 (PR #1558). The editor did not work in `session` mode because the editor tab was SFTP-shaped: `EditorTabMeta` could only reference an `SftpManager` session. FTP, Docker and agent-session file browsers therefore had a dead Edit action — this is why #1335's "previews" acceptance criterion was never fully met.

## What changed

**`EditorTabMeta` gained a seam rather than being reshaped wholesale.** `isRemote` keeps meaning "not on the local disk"; which remote transport backs the tab is now decided by exactly one of two mutually exclusive fields:

- `sftpSessionId` — the legacy `SftpManager`/`sftp_*` path, used by SSH.
- `sessionBrowser: { sessionId, connectionType }` — the protocol-agnostic session layer, used by every other file-browser-capable type.

This turned out to be the cheap change, because the existing code was already written the right way. Every SFTP-specific affordance — writability probe, exec-capability probe, elevated/sudo writes, realpath, download, "save a copy" — already gates on `!!meta.sftpSessionId`, not on `isRemote`. So they stay off for session-backed tabs with **no new gating code**, and there was no need to duplicate a layer that exists: `sessionReadFile`/`sessionWriteFile` and the `session_*` command surface were already in place from #1335, so this PR adds no backend code at all.

Concretely:

- `src/types/terminal.ts` — new `EditorSessionRef`; `sessionBrowser` on `EditorTabMeta`, with the two-transport contract documented.
- `src/store/appStore.ts` — `openEditorTab` takes an optional `sessionBrowser`. On refresh of an existing tab, only the supplied transport is set and the other is cleared, so a tab can never carry both (which would let the SFTP-first branch shadow the session one).
- `src/components/FileEditor/FileEditor.tsx` — load and save branch to `session_read_file` / `session_write_file`. These are byte-oriented (they back binary transfers too), so the editor decodes/encodes UTF-8 at the boundary; the SFTP commands already hand back a string.
- `src/components/Sidebar/FileBrowser.tsx` — the `edit` action works in `session` mode, and the editor-tab branch of the mode effect routes a session-backed tab to `session` instead of falling into `sftp` (which would have pointed the browser at an `SftpManager` session that does not exist).

Worth noting: the Edit context item, double-click, and Enter were never mode-gated — they always called the action, which silently no-op'd in session mode. All three paths now work; no new testids, so the catalog is unchanged.

## Tests

`FileEditor.test.tsx` — reads via `session_read_file` and never touches `sftp_*`; decodes non-ASCII bytes; saves via `session_write_file` with the right session/path/bytes; surfaces a failed save and keeps the buffer dirty; surfaces a read failure as a load error; and a guard that a session tab is badged remote but offers **no** sudo / save-a-copy / download / read-only badge even when the SFTP probes would answer affirmatively.

`FileBrowser.test.tsx` — a session-backed editor tab lands in `session` mode with the right session id; SFTP and local editor tabs are unchanged; and activating a file in session mode opens an editor tab carrying `sessionBrowser` and no `sftpSessionId`.

**Verified red without the implementation**: 5 of 6 `FileEditor` cases and 2 of 4 `FileBrowser` cases fail on the pre-change code. The rest are deliberate regression guards for the SFTP/local paths, and I am not claiming they prove the new behavior.

## Test plan

- `npx tsc --noEmit` — clean.
- `npx vitest run` — 3286 passed / 351 files.
- `./scripts/check.sh` — all checks pass (prettier caught one file; fixed in its own commit).
- `cargo test --workspace --all-features` — all pass except `termihub-core --test docker_spawn`, which **fails identically on clean `develop` (803ed117)** on this host and is unrelated: this PR is frontend-only and touches zero Rust.

Closes #1557

### DIFF
--- a/docs/changes/dev2/feature/1557-session-layer-editor.md
+++ b/docs/changes/dev2/feature/1557-session-layer-editor.md
@@ -1,0 +1,13 @@
+### Added
+
+- File editor support for session-layer file browsing: files opened from an FTP,
+  Docker, or agent-session file browser now open in the editor and can be saved
+  back over the session layer. Previously the Edit action (and double-click /
+  Enter on a file row) silently did nothing for these connection types, because
+  the editor tab could only reference an SSH `SftpManager` session. An editor tab
+  can now reference a session-layer browser instead, reading and writing through
+  `session_read_file` / `session_write_file`, and a session-backed editor tab no
+  longer puts the file browser into SFTP mode. The SFTP-specific affordances
+  (read-only detection, sudo/elevated writes, "save a copy", download) remain
+  exclusive to SSH, which the session layer does not expose. SFTP editing is
+  unchanged (#1557).
--- a/src/components/FileEditor/FileEditor.test.tsx
+++ b/src/components/FileEditor/FileEditor.test.tsx
@@ -814,3 +814,151 @@ describe("FileEditor — SFTP-only read-only fallback (#1330)", () => {
     expect(query("file-editor-download")).toBeNull();
   });
 });
+
+describe("FileEditor — session-layer backed tabs (#1557)", () => {
+  const SESSION_META: EditorTabMeta = {
+    filePath: "/srv/app/config.yml",
+    isRemote: true,
+    sessionBrowser: { sessionId: "sess-ftp-1", connectionType: "ftp" },
+  };
+
+  beforeEach(() => {
+    (globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = true;
+    container = document.createElement("div");
+    document.body.appendChild(container);
+    root = createRoot(container);
+    useAppStore.setState({ ...useAppStore.getInitialState() });
+    vi.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    act(() => root.unmount());
+    container.remove();
+    vi.clearAllMocks();
+    (globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT = false;
+  });
+
+  /** Encode text the way session_read_file returns it: a plain byte array. */
+  function bytes(text: string): number[] {
+    return Array.from(new TextEncoder().encode(text));
+  }
+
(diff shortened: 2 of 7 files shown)


armaxri:
Filed #1599 as a follow-up: editor tabs dedupe on `filePath` + `isRemote` only, so the same path opened from two different remote sessions collides into one tab. Pre-existing (two SSH hosts always collided), but this PR widens the blast radius by letting FTP/Docker/agent sessions open editor tabs too.

Issue #1573, opened 2026-07-17
fix(ui): Transfer Queue row for a paste-upload is named after the temp file


armaxri:
Found while implementing #1532 (PR #1565), which asserts the current behaviour so the row names stay pinned.

## What

An SFTP→SFTP copy/paste is implemented as "download to a local temp file, re-upload to the destination" (`useFileSystem.pasteEntry`):

```
const tempPath = `/tmp/termihub-paste-${Date.now()}-${clipEntry.name}`;
await sftpDownload(clipboard.sftpSessionId, clipEntry.path, tempPath);
await sftpUpload(sftpSessionId, tempPath, destPath);
```

The backend derives a transfer's `file_name` from the **remote** path on download but the **local** path on upload (`src-tauri/src/commands/files.rs`, `file_name_of`):

- `sftp_download` → `file_name_of(&remote_path)`
- `sftp_upload` → `file_name_of(&local_path)`

So pasting `report.csv` shows two Transfer Queue rows:

| direction | name shown | path shown |
| --- | --- | --- |
| download | `report.csv` | `/home/testuser/report.csv` |
| upload | `termihub-paste-1784278708447-report.csv` | `/home/testuser/dest/report.csv` |

Verified live against the `ssh-password` fixture.

## Why it matters

The upload row exposes an internal temp filename with a timestamp to the user for a file they know as `report.csv`. The `path` column (added by #1531) is already correct — only the name cell is wrong.

## Options

- Have `sftp_upload` prefer the destination remote path's basename for `file_name` (it already receives `remote_path`, and that is what `path` is populated from). Would need a check that no other upload caller relies on the local basename.
- Or let the paste caller pass an explicit display name.

Note #1565's live suite asserts the *current* behaviour (`name.startswith("termihub-paste-")`); that assertion should be updated to expect the clean name when this is fixed.

PR #1611, opened 2026-07-17
fix(backend): name ftp_upload's transfer row from the remote path (Closes #1594)

### LINKED ISSUE #1594: fix(backend): ftp_upload names its transfer row from the local path, like sftp_upload did
Follow-up to #1573 (PR pending).

## What

#1573 fixed `sftp_upload` deriving a Transfer Queue row's `file_name` from the **local** path while its `path` cell comes from the **remote** path — so the two cells could describe different files. The rule established there: a transfer row's name is the base name of the path it displays.

`ftp_upload` (`src-tauri/src/commands/transfer.rs:143`) still has the old shape:

```rust
let file_name = file_name_of(&local_path);
let handle = registry.enqueue(&transfer_id, &session_id, TransferDirection::Upload, &file_name, &remote_path, 0);
```

## Why this was left out of #1573

It is **latent, not live**. The SFTP bug was reachable because `useFileSystem.pasteEntry` implements SFTP→SFTP copy as download-to-temp then re-upload, so the local file is termiHub's scratch copy (`/tmp/termihub-paste-<ts>-<name>`). No FTP caller does that today, so no current FTP path produces a mismatched name. #1573 was kept to the reported bug rather than changing a feature-gated path it had no fixture to verify.

## Scope

- [ ] Decide whether `ftp_upload` should adopt the same rule (`file_name_of(&remote_path)`) for consistency.
- [ ] Note that `file_name_of` is duplicated in `commands/files.rs` and `commands/transfer.rs` (the latter `#[cfg(feature = "ftp")]`-gated) — worth sharing one helper if both converge on the same rule.
- [ ] Only meaningful with FTP built in (`--features ftp`).

## Done when

`ftp_upload` and `sftp_upload` agree on where a row's display name comes from, or there is a recorded reason they should differ.

armaxri:
## What

`ftp_upload` (`src-tauri/src/commands/transfer.rs`) derived the Transfer Queue row's `file_name` from the **local** path, while its `path` cell comes from the **remote** path — so the two cells could describe different files. This is the same defect that #1573 (PR #1595) fixed for `sftp_upload`.

The rule #1573 established: **a transfer row's name is the base name of the path it displays.** This PR applies that invariant to `ftp_upload` — a one-line change at the single `enqueue` site, `file_name_of(&local_path)` → `file_name_of(&remote_path)`, exactly mirroring the SFTP fix.

## Latent, not live — the honest case stays a no-op

This was deliberately left out of #1573 because it is **latent**: no FTP caller does download-to-temp-then-reupload today (unlike `useFileSystem.pasteEntry` for SFTP), so no current FTP path produces a mismatched name. For a genuine local→remote FTP upload the local and remote basenames are equal, so naming from either is identical — this change is **byte-for-byte unchanged** for every honest upload. It only differs where a future caller uploads from a scratch/temp local file the user never named, which is precisely the case where naming from the remote path keeps the scratch name off the row.

## Tests

Added `ftp`-gated Rust unit tests for `file_name_of` in `transfer.rs`, mirroring #1573's unit tests. These **do** run in CI under `cargo test --workspace --all-features`, unlike the system suite which is `-m "not integration"`-excluded (#1569).

## Verification — and its limit (stated plainly)

- `cargo test -p termihub --features ftp` — the 3 new tests pass.
- `cargo fmt --all --check` and `cargo clippy -p termihub --features ftp --all-targets -- -D warnings` are clean.

**What I could not verify:** there is no FTP fixture that drives `ftp_upload`'s transfer-row naming end-to-end. The FTP system tests (`test_transfer_checkers.py`) exercise the harness-side `ftp_download` checker via `pyftpdlib`, not the Rust command's `enqueue` naming. This is the same gap #1573's author flagged as the reason they deferred this. Consequently the added unit tests guard the `file_name_of` invariant but — because the helper itself is unchanged — do not by themselves fail on a regression of the call site; the call-site change is verified by code review against the identical, already-merged SFTP fix, not by a red test run. No live FTP verification is claimed.

Refs the SFTP precedent: #1573 / PR #1595 (merge `65c8a3b2`).

Closes #1594

### DIFF
--- a/src-tauri/src/commands/transfer.rs
+++ b/src-tauri/src/commands/transfer.rs
@@ -140,7 +140,13 @@ pub async fn ftp_upload(
         debug!(session_id, local_path, remote_path, "FTP upload");
         let config = parse_ftp_config(config)?;
         let transfer_id = uuid::Uuid::new_v4().to_string();
-        let file_name = file_name_of(&local_path);
+        // Named for the *remote* path, not the local one, so the name always
+        // agrees with the `path` this row displays (#1594, mirroring the SFTP
+        // fix in #1573). Every honest local→remote upload builds `remote_path`
+        // as `<dir>/<basename of local>`, so this is byte-for-byte unchanged
+        // for them; it only differs for a future caller that uploads from a
+        // scratch/temp path the user never named (as SFTP→SFTP paste does).
+        let file_name = file_name_of(&remote_path);
         let handle = registry.enqueue(
             &transfer_id,
             &session_id,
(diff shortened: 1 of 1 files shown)


armaxri:
Follow-up filed: #1612 — dedupe the now-identical `file_name_of` helper shared between `commands/files.rs` and `commands/transfer.rs` (called out in #1594's Scope).

PR #1631, opened 2026-07-17
test(sftp): sync live transfer-queue test on the store slice (Closes #1619)

### LINKED ISSUE #1619: test(sftp): flaky 'Transfer Queue panel' timeout in test_real_sftp_paste_populates_the_transfer_queue
Follow-up to #1593 (PR #1618).

## What

While validating #1593 with repeated **local** live runs of `TestTransferQueueLiveTransfer.test_real_sftp_paste_populates_the_transfer_queue` (`tests/system/tests/test_transfer_queue.py`), a **separate** flake surfaced — distinct from the password-prompt race #1593 fixed.

After the paste click, the test waits for the Transfer Queue panel to appear:

```python
self.driver.click("file-browser-paste")
# The panel appears as soon as the first transfer registers.
self.wait(lambda: self.driver.exists(QUEUE), what="the Transfer Queue panel")  # line 203
```

Intermittently this times out (default 20s):

```
AssertionError: timed out waiting for the Transfer Queue panel (last error: None)
```

The backend logs show the SFTP session connected and the download/upload actually ran, so the transfer registers but the `QUEUE` panel element does not become visible within the wait budget. This is **not** the #1593 password-prompt race (that produced `no element with data-testid="password-prompt-connect"`, and was eliminated) — the failure occurs later, after SFTP is connected.

## Reproduction

~5 failures across ~35 consecutive local runs (debug build, macOS, real `ssh-password` container). Frequency was higher under rapid back-to-back runs, suggesting a timing/rendering race in the panel opening rather than a hard bug. Not seen in CI because the integration lane is excluded (#1569).

## Scope

- [ ] Determine why the Transfer Queue panel (`QUEUE` testid) can fail to appear within the wait budget after the first transfer registers — is it a genuine panel-open race, a slow first-render under load, or too tight a timeout?
- [ ] Synchronise the test on the real readiness signal (e.g. the `transferQueue` store slice gaining its first entry) rather than only the panel DOM, or fix the panel-open path if it is a real race.
- [ ] Avoid papering over it with a bumped timeout unless that is demonstrably correct.

## Done when

Repeated local runs of `test_real_sftp_paste_populates_the_transfer_queue` no longer flake on the Transfer Queue panel failing to appear.

armaxri:
## What

`TestTransferQueueLiveTransfer.test_real_sftp_paste_populates_the_transfer_queue` waited on the panel **DOM** (`exists(transfer-queue)`) as its first checkpoint after the paste click. That raced the render and could time out with *"timed out waiting for the Transfer Queue panel"* even though the transfer had already registered — the flake reported in #1619.

This waits on the `transferQueue` **store slice** gaining its first entry instead — the authoritative, render-independent readiness signal, and the one every subsequent assertion already depends on.

Closes #1619.

## Why this is the right signal (not a papered-over timeout)

- The panel (`TransferQueue`) is mounted **unconditionally** in `App.tsx`; it returns `null` purely while `entries.length === 0`. `transferQueueMinimized` defaults `false` and is not persisted, so it never gates a fresh app.
- `applyTransferProgressToQueue` folds the first `transfer-progress` event into the store slice **synchronously in the event callback** (`useTransferEvents`). The panel's `data-testid` only paints on the **following React commit**, which trails the store.
- So "a transfer has registered" (store slice non-empty) is the true readiness condition; the DOM is a lagging derivative of it. The change moves only the *synchronisation point* — the panel's own rendering is still asserted afterwards (`transfer-row` / summary reads after the two-row completion wait), so no coverage is dropped. No sleep, no bumped timeout.

## Evidence (local hammering on this machine)

The integration lane is CI-dark (#1569), so this was validated locally against the live `ssh-password` Docker fixture, debug build:

- **60 runs of the live test** (30 idle + 30 under 8x CPU oversubscription, load avg ~21): **0 flakes** before and after.
- **30 instrumented runs** measuring the gap between the store slice gaining its entry and the panel DOM appearing: **<30 ms in every run** (mostly <3 ms; occasionally the DOM sampled first). The panel-open path itself is sound — there is no multi-second render race on an idle/CPU-bound machine.
- Post-change: **6/6** live runs green.

## Scope note / follow-up

The original report saw ~5/35 failures during a multi-slot round under **memory** pressure. The measured store-vs-DOM gap (<30 ms) is far too small to explain a 20 s timeout by render lag alone, which points at a residual **event-delivery** gap under memory pressure (the store slice never receiving the event within budget) rather than the panel code. That is out of scope for this test change and is filed as a follow-up. This change makes the test synchronise on the correct signal and, if that residual gap ever bites, fail with a message that names it (*"the transfer queue to register the first transfer"*) instead of misdirecting at the panel.

## Test plan

- `./scripts/test-system-py.sh --debug --fixtures "ssh-password" -m integration -k test_real_sftp_paste_populates_the_transfer_queue` — green across repeated runs.

### DIFF
--- a/tests/system/tests/test_transfer_queue.py
+++ b/tests/system/tests/test_transfer_queue.py
@@ -199,8 +199,22 @@ class TestTransferQueueLiveTransfer(
         )
         self.driver.click("file-browser-paste")
 
-        # The panel appears as soon as the first transfer registers.
-        self.wait(lambda: self.driver.exists(QUEUE), what="the Transfer Queue panel")
+        # Synchronise on the store slice — the authoritative readiness signal —
+        # not the panel DOM. `applyTransferProgressToQueue` folds the first
+        # `transfer-progress` event into the `transferQueue` slice synchronously
+        # in the event callback (`useTransferEvents`), and the panel is mounted
+        # unconditionally, gated only by that slice being non-empty
+        # (`TransferQueue` returns `null` while `entries.length === 0`). So "a
+        # transfer has registered" is the real readiness condition; the panel's
+        # `data-testid` only paints on the following React commit, which trails
+        # the store, so waiting on the DOM here raced (#1619). The panel's own
+        # rendering is still asserted below (the `transfer-row` / summary reads
+        # after the two-row wait), so this only moves the *synchronisation* onto
+        # the earlier, render-independent signal — it does not drop coverage.
+        self.wait(
+            lambda: self.queue_entries() or False,
+            what="the transfer queue to register the first transfer",
+        )
 
         # Download + upload = two rows, both settling at `completed`.
         entries = self.wait(


armaxri:
Follow-up filed: #1632 (Transfer Queue can miss a transfer-progress event under memory pressure).

Issue #1632, opened 2026-07-17
Transfer Queue can miss a transfer-progress event under memory pressure (panel never opens)


armaxri:
Follow-up to #1619 (PR #1631).

## Context

#1619 was reported as a flaky *"timed out waiting for the Transfer Queue panel"* in the live SFTP paste test during a multi-slot round under memory pressure (~5/35). PR #1631 hardened that **test** to synchronise on the `transferQueue` store slice (the authoritative, render-independent readiness signal) instead of the panel DOM.

While doing so, instrumentation showed the panel-open path is **sound**: across 30 live runs under 8x CPU oversubscription, the gap between the store slice gaining its first entry and the panel DOM painting was **<30 ms** (mostly <3 ms). A ~20 s timeout cannot be explained by render lag of that size.

## Suspicion

That leaves the **frontend not folding any `transfer-progress` event** into the `transferQueue` slice within budget, despite the backend logs showing the transfer ran and emitted events. The original failure showed `last error: None` (a responsive bridge), so the webview was alive — the `listen("transfer-progress")` subscription in `useTransferEvents` simply did not observe the event(s) for that transfer. The most likely trigger is the severe memory pressure of a multi-slot round (jetsam), which could delay/drop Tauri event delivery to the webview.

This was **not reproduced** on a RAM-ample machine (60 live runs, 30 idle + 30 under heavy CPU load, 0 failures) — it appears to need real memory pressure, which is why it only showed during a busy multi-slot round.

If confirmed, this is a genuine (if rare) user-facing risk: a user finishes a transfer under memory pressure and the Transfer Queue panel that should show it never appears.

## Scope

- [ ] Reproduce under real memory pressure (not just CPU) — e.g. drive the live paste test while the machine is held near the jetsam threshold.
- [ ] Determine whether the `transfer-progress` event is genuinely not delivered to the webview `listen` subscription, or delivered but not folded, when the panel fails to open.
- [ ] If events can be dropped/delayed, add resilience: e.g. reconcile the `transferQueue` slice from a backend snapshot on a suitable signal, or make the emit path more robust under pressure.

## Done when

The Transfer Queue panel reliably opens after a transfer registers even under memory pressure, or the delivery gap is shown not to exist and #1619's residual risk is documented as environmental.

PR #1653, opened 2026-07-18
fix(sftp): reconcile transfer queue backstop for a dropped terminal event (Closes #1645)

### LINKED ISSUE #1645: Transfer Queue: reconcile backstop for a dropped terminal transfer event
Follow-up to #1632 (PR #1644).

#1644 makes the Transfer Queue **panel open** independent of `transfer-progress` event delivery by seeding a `queued` row from the transfer's registration id. That fixes the reported "panel never opens" symptom. It does **not** self-heal a row whose *terminal* event (`done`/`error`/`cancelled`) is also dropped under memory pressure: such a row stays visually stuck at `queued`/`active`.

A backend-snapshot reconcile (the shape #1632 originally suggested) is the natural backstop, but today it cannot cover the reported SFTP case:

- SFTP transfers register in the registry's **legacy token map**, not the **rich** map that `transfer_list` snapshots — so they never appear in `transfer_list`.
- Rich transfers are `drop_entry`'d on completion, so a *completed* transfer is gone from the snapshot.

## Scope

- [ ] Retain recently-terminal transfers in `TransferRegistry` (bounded history) so a completed/failed/cancelled transfer is still snapshot-able for a short window.
- [ ] Include in-flight (and recently-terminal) **legacy SFTP** transfers in `transfer_list` — carry minimal metadata (session, direction, file name, path, transferred/total, terminal state) on the legacy registration.
- [ ] Add a frontend reconcile that calls `transfer_list()` on a suitable signal (e.g. a periodic tick while rows are non-terminal, or on window focus) and upserts the snapshot into `transferQueue`, so a stuck seeded row settles to its true final state.

## Done when

A transfer whose **every** `transfer-progress` event (including the terminal one) is dropped still ends up showing its correct terminal state in the Transfer Queue within a bounded reconcile window, for both SFTP (legacy) and FTP (rich) transfers.

armaxri:
## What & why

Follow-up to #1632 (PR #1644). #1644 makes the Transfer Queue **panel open**
independent of `transfer-progress` delivery by seeding a `queued` row from the
transfer's registration id. It does **not** self-heal a row whose *terminal*
event (`done`/`error`/`cancelled`) is also dropped under memory pressure: such a
row stays visually stuck at `queued`/`active` forever.

This adds the reconcile backstop #1645 asks for. A naive "reconcile from
`transfer_list`" could not work before, for two reasons #1632 documented:

- SFTP transfers register in the registry's **legacy token map**, not the
  **rich** map that `transfer_list` snapshots — so they never appeared at all.
- Rich transfers are `drop_entry`'d on completion — so a *completed* rich
  transfer was gone from the snapshot too.

Both are fixed here, then the frontend reconciles against the snapshot.

## Backend — make the terminal state observable

- **Legacy SFTP now appears in `transfer_list`.** `register` carries the
  transfer identity (session, direction, file name, path, total); a live legacy
  transfer lists as `active`.
- **Recently-terminal transfers are retained** in a bounded, TTL'd history
  (**60 s / 64 entries**) instead of vanishing on completion. `finish_legacy`
  records the SFTP terminal state (`completed`/`cancelled`/`failed` + final byte
  count); `drop_entry` records a completed/cancelled **rich** transfer's
  snapshot before removing it. `transfer_list` returns live rich + live legacy +
  retained-terminal, with live entries shadowing retained ones by id.

### Retention policy & why

The frontend polls on a few-second tick while any row is in-flight, so a **60 s
TTL** is an order of magnitude larger than the catch window — ample to settle a
transfer whose terminal event was dropped, without holding completed transfers
around. The **64-entry cap** bounds memory against a completion burst between
TTL sweeps (each retained entry is a small `TransferSnapshot`). Both bounds
evict oldest-first, so memory cannot leak. Transfer ids are UUIDv4, so id reuse
is a non-issue; `list()` guards it anyway by letting a live entry shadow a
retained snapshot of the same id.

## Frontend — the reconcile

`useTransferReconcile` polls the reliable request/response `transfer_list`
snapshot **while any row is non-terminal** (and on window focus) and settles a
stuck row to its true terminal state via a new `reconcileTransferQueue` store
action. It is conservative and idempotent:

- Only settles a still-open row to a **terminal** snapshot state — events own
  live progress, so a lagging/zero snapshot can never regress a live row.
- Never resurrects a removed row; never clobbers an already-terminal row.
- Polling is gated on a pending row, so an idle app makes **no** backend calls.

## Reproduction / verification

Real jetsam can't be manufactured on a shared multi-slot machine (and must not
be driven to OOM), so the fix is proven by **simulating a fully-dropped event
stream** — including the terminal event — and asserting the reconcile settles
the row from the backend snapshot alone:

- `useTransferReconcile.test.ts` — a seeded row whose every `transfer-progress`
  event was dropped settles to `completed` from the `transfer_list` snapshot;
  no backend call is made when nothing is pending.
- `appStore.transferQueue.test.ts` — `reconcileTransferQueue` settles a stuck
  seeded/active row; is idempotent; never clobbers a settled row; never moves a
  live row; never resurrects a removed row.
- `transfer.test.ts` — `transferEntryFromSnapshot` mapping.
- `registry.rs` — legacy transfer lists as active; `finish_legacy` / rich
  `drop_entry` retain terminal snapshots after the live entry is gone; capacity
  and TTL eviction; live shadows retained; session filtering.

`tsc --noEmit`, `pnpm build`, eslint, `cargo fmt`/`clippy -D warnings`, the full
frontend suite (3390) and the transfer backend tests are green.

## Follow-up

Two low-severity notes from self-review (not blocking, both self-correcting):
a rich FTP transfer transiently `failed` mid-retry could be settled to `failed`
by a reconcile; and the pending-row selector re-scans on every store mutation.
Will file a `Ready2Implement` follow-up referencing #1645 if worth hardening.

Closes #1645

### DIFF
--- a/docs/changes/dev5/fix/1645-transfer-terminal-reconcile.md
+++ b/docs/changes/dev5/fix/1645-transfer-terminal-reconcile.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- Transfer Queue: a transfer now settles to its correct final state even when
+  its terminal `transfer-progress` event is dropped (e.g. under memory
+  pressure). Previously a row seeded at registration (#1632) could stay stuck at
+  `queued`/`active` forever if the `done`/`error`/`cancelled` event never
+  arrived. The backend now retains recently-terminal transfers briefly and
+  includes legacy SFTP transfers in the `transfer_list` snapshot, and the
+  frontend reconciles the queue against that snapshot while any transfer is
+  in-flight (and on window focus), settling stuck rows without ever regressing a
+  live one (#1645).
--- a/src-tauri/src/commands/files.rs
+++ b/src-tauri/src/commands/files.rs
@@ -162,12 +162,20 @@ pub async fn sftp_download(
     let (dedicated, total) = open_transfer_channel(session, Some(remote_path.clone())).await?;
 
     let transfer_id = uuid::Uuid::new_v4().to_string();
-    let token = registry.register(&transfer_id);
+    let file_name = file_name_of(&remote_path);
+    let token = registry.register(
+        &transfer_id,
+        &session_id,
+        TransferDirection::Download,
+        &file_name,
+        &remote_path,
+        total,
+    );
     let ctx = TransferContext {
         transfer_id: transfer_id.clone(),
         session_id,
         direction: TransferDirection::Download,
-        file_name: file_name_of(&remote_path),
+        file_name,
         path: remote_path.clone(),
         total,
     };
@@ -211,7 +219,15 @@ pub async fn sftp_upload(
         .unwrap_or(0);
 
     let transfer_id = uuid::Uuid::new_v4().to_string();
-    let token = registry.register(&transfer_id);
+    let file_name = file_name_of(&remote_path);
+    let token = registry.register(
+        &transfer_id,
+        &session_id,
+        TransferDirection::Upload,
+        &file_name,
+        &remote_path,
+        total,
+    );
     let ctx = TransferContext {
         transfer_id: transfer_id.clone(),
         session_id,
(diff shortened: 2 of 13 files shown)


armaxri:
Follow-up filed: #1657 (harden reconcile against transient rich-`failed` state; memoize the pending-row selector).

Issue #1704, opened 2026-07-19
feat(terminal): per-connection highlighting override and status-bar quick toggle


armaxri:
Two additional entry points for controlling highlighting: a per-connection override and a status-bar quick toggle. Part of #1696; concept: `docs/concepts/partial/terminal-syntax-highlighting.md` (#522) → "Per-Connection Override" and "Status Bar Indicator".

## Scope

- **Per-connection override** — `src/components/ConnectionEditor/ConnectionTerminalSettings.tsx`: add a `Use global default / Always on / Always off` dropdown writing `TerminalOptions.syntaxHighlighting.override` (type from #1, field from #2). Optional connection-specific additional rules (append after global rules; cannot remove global rules) per concept — may be deferred to a follow-up if it grows the PR.
- **Status-bar quick toggle** — `src/components/StatusBar/StatusBar.tsx`: show a "Highlighting: ON/OFF" indicator for the active session when the feature is enabled; clicking flips the **per-session temporary toggle** (`setSessionHighlighting`, non-persisted; from #2) that the engine (wired in #3) already reacts to.

## Shared files / collision surface

- **`src/components/StatusBar/StatusBar.tsx`** — not a macro-collision file.
- **`src/store/appStore.ts`** — if any store action beyond #2's `setSessionHighlighting` is needed, this touches appStore, which the **macro chain (#1675, #1670)** also edits. Prefer to consume the actions already added in #2; if new appStore edits are required, serialize against macro rounds.
- `ConnectionTerminalSettings.tsx` — not a macro-collision file.

## Dependencies / landing order

After #2 (override field + session toggle state) and #3 (engine reacts to both). Independent of #4/#5.

## Acceptance criteria

- Per-connection `Always off` disables highlighting for that connection even when global is on (and vice versa); `Use global default` follows global.
- Status-bar click toggles the current session live without persisting; reconnect/reopen returns to the resolved default.
- Component tests for the dropdown resolution and the status-bar toggle. `./scripts/test.sh` + `./scripts/check.sh` green. No `any`.

Refs #522, #1696.

Issue #1700, opened 2026-07-19
feat(terminal): wire syntax highlighting engine into the terminal lifecycle


armaxri:
Instantiate and drive the `SyntaxHighlightingEngine` (child #1) from the live terminal, gated by the store config (child #2). Part of #1696; concept: `docs/concepts/partial/terminal-syntax-highlighting.md` (#522).

## Scope

- **`src/components/Terminal/Terminal.tsx`**: in the terminal-creation effect, after `xterm.open(...)`, construct a `SyntaxHighlightingEngine(xterm)`. Resolve the effective config via `resolveHighlightingConfig(appSettings.syntaxHighlighting, tabTermOpts?.syntaxHighlighting)` and `enable(resolveActiveRules(...))` when enabled. Add a settings-reactive effect (mirroring the existing font/theme/scrollback effect around lines 1231-1254) that calls `updateRules` / `enable` / `disable` when the global config, the per-connection override, or the per-session toggle changes. Dispose the engine in the effect cleanup (alongside the existing xterm teardown).
- Keep the wiring **minimal and behind the config** so it ships safely even before any UI exists (feature controllable only via store defaults until child #4).
- Respect the reconnect/reattach and StrictMode teardown patterns already in this file — the engine must be created/disposed on the same boundaries as the xterm instance so a reconnect (retryCount bump) rebuilds it cleanly.

## Shared files / collision surface

- **`src/components/Terminal/Terminal.tsx`** — **HIGH-COLLISION. The macro chain (#1670 / #1675 macro playback / #1676 / #1677) edits the Terminal component area.** Do NOT run this in the same round as any macro issue — serialize. Landing this child quickly closes the Terminal.tsx collision window for the rest of this epic.
- Reads store state added in #2; no other shared-file edits.

## Dependencies / landing order

After #1 (engine) and #2 (store/config). Blocks #4 and #6 (they need visible engine behavior). This is the last child that touches Terminal.tsx — land it before scheduling later children against macro rounds.

## Acceptance criteria

- With the feature enabled in settings, plain output (e.g. `cat` of a log with `ERROR`/`WARN`/an IP/a URL) shows the expected recoloring; server-ANSI-colored output is untouched.
- Toggling the config off removes all decorations live; disposing the tab leaks no decorations or listeners.
- Alt-screen apps (vim/less/htop) are unaffected (engine self-disables on alt buffer).
[...]

Issue #1697, opened 2026-07-19
feat(terminal): syntax highlighting engine core, built-in rules, and types


armaxri:
Foundation for terminal output syntax highlighting. Part of #1696; concept: `docs/concepts/partial/terminal-syntax-highlighting.md` (#522).

This is a **self-contained, unit-testable module** — no wiring into the live terminal yet, so it can be built and reviewed in isolation.

## Scope

- **`src/types/syntaxHighlighting.ts`** (new): `HighlightStyle`, `HighlightRule`, `SyntaxHighlightingConfig`, `ConnectionHighlightingOverride`, `ConnectionHighlightingConfig` (see concept "Type Definitions").
- **`src/services/syntaxHighlightingRules.ts`** (new): the built-in rule set (error/warning/success keywords, URLs, file paths, IP addresses, numbers as P0/P1 default-on; quoted strings, emails, MAC, dates, UUIDs, hex as P2/P3 default-off). Theme-aware color resolution (`getThemedRuleColor`) resolving to `#RRGGBB` — decoration `foregroundColor` only accepts `#RRGGBB`.
- **`src/services/syntaxHighlighting.ts`** (new): `SyntaxHighlightingEngine` class bound to an xterm instance:
  - compile the active rules into a matcher; recompile on `updateRules`
  - `onWriteParsed`-driven dirty-line scanning; per-line text extraction honoring `isWrapped` (treat wrapped rows as one logical line)
  - overlap resolution: longest match wins, then priority, then first-match
  - **ANSI-conflict avoidance**: for each candidate cell use `IBufferCell.isFgDefault()`; only decorate default-fg cells; split a match around already-colored cells
  - apply via `registerMarker` + `registerDecoration({ marker, x, width, foregroundColor, ... })` — **native `foregroundColor` option, not `onRender` element styling**
  - **performance primitives** (engine-internal): dirty-line tracking, per-frame line budget, throughput throttle (pause when writes/sec high, backfill when settled), skip lines > ~10k chars, disable while `buffer.active.type === 'alternate'`, discard decorations on screen clear/reset
  - `enable` / `disable` / `updateRules` / `dispose` with full decoration cleanup

## Shared files / collision surface

**None.** All three files are new; this child touches no file the macro chain edits. Safe to run in parallel with anything, including macro work.

## Dependencies / landing order

First. Blocks #1696 children 2–6.

## Acceptance criteria

[...]

Issue #1696, opened 2026-07-19
[Epic] Terminal output syntax highlighting


armaxri:
Tracking epic for client-side syntax highlighting of terminal output, per the concept in `docs/concepts/partial/terminal-syntax-highlighting.md` (concept issue #522).

## Goal

Colorize recognizable patterns in **plain, uncolored** terminal output — error/warning/success keywords, URLs, file paths, IP addresses, numbers, etc. — without ever overriding colors the remote server already set via ANSI SGR, and without touching the PTY byte stream.

## Feasibility (verified against bundled xterm.js v6.0.0)

Implementable with the currently bundled xterm.js. Verified APIs:

- **Recoloring**: `IDecorationOptions.foregroundColor` / `backgroundColor` (xterm v6) recolor the actual buffer cell(s) — no byte-stream rewriting needed. Use these native options, **not** the `decoration.onRender(el => el.style.color = ...)` approach sketched in the concept — the native option is what recolors the glyph rather than an empty overlay div. (`#RRGGBB` only.)
- **ANSI-conflict avoidance**: `IBufferCell.isFgDefault()` lets the engine skip cells whose foreground the server set, so server ANSI colors always win.
- **Write hook**: `Terminal.onWriteParsed`. **Alt-screen guard**: `buffer.active.type === 'alternate'`.

The core PTY/xterm data path is **not** modified — highlighting is a purely additive, post-parse decoration layer.

## Collision surface (coordinator: read before scheduling)

Some children edit files the **macro chain (#1670 / #1675 / #1676 / #1677)** also touches. Do not run those in the same round as macro work:

- **`src/components/Terminal/Terminal.tsx`** — #1700. Serialize against macro work.
- **`src/store/appStore.ts`** — #1698 and (possibly) #1704. Serialize against macro playback (#1675).

The foundation (#1697) and UI-only children (#1701, #1703) touch **no** macro-collision files and can run in parallel with macro work.

## Children (land in this order)

1. #1697 — engine core + built-in rules + types (no shared files)
2. #1698 — settings model + store state (appStore.ts, types/connection.ts)
3. #1700 — wire engine into terminal lifecycle (Terminal.tsx)
4. #1701 — global settings UI (built-in rule toggles)
5. #1703 — custom highlight rules editor + regex safety
6. #1704 — per-connection override + status-bar quick toggle

Refs #522.

armaxri:
[...]

PR #1734, opened 2026-07-19
feat(terminal): add syntax highlighting settings model and store state

### LINKED ISSUE #1698: feat(terminal): syntax highlighting settings model and store state
Persisted config + store state for terminal syntax highlighting. Part of #1696; concept: `docs/concepts/partial/terminal-syntax-highlighting.md` (#522).

## Scope

- **`src/types/connection.ts`**: extend `AppSettings` with optional `syntaxHighlighting?: SyntaxHighlightingConfig` and `TerminalOptions` with optional `syntaxHighlighting?: ConnectionHighlightingConfig` (types from child #1 / `src/types/syntaxHighlighting.ts`).
- **`src/store/appStore.ts`**: global `syntaxHighlighting` config in settings (default: P0/P1 rules on, P2/P3 off, no custom rules, feature enabled or disabled per product default), plus **per-session temporary toggle** state + action (e.g. `sessionHighlighting: Record<sessionId, boolean>` / `setSessionHighlighting`) used by the status-bar quick toggle (child #6). Include a `resolveHighlightingConfig(global, perConnection)` + `resolveActiveRules(config)` helper (co-locate in a util or the service from #1).
- Migration/back-compat: absent config resolves to the built-in defaults.

## Shared files / collision surface

- **`src/store/appStore.ts`** — **also edited by the macro chain (#1675 macro playback, #1670).** Do NOT schedule this in the same round as macro appStore work; serialize.
- `src/types/connection.ts` — not a macro-collision file.

## Dependencies / landing order

After #1 (needs the config/rule types). Blocks children 3–6.

## Acceptance criteria

- Unit tests for `resolveHighlightingConfig` (global on/off, `always-on`, `always-off`, `global`) and default fallback.
- Store persists global config and round-trips through workspace/settings save-restore; per-session toggle is in-memory only (not persisted), per concept.
- `./scripts/test.sh` and `./scripts/check.sh` green. No `any`.

Refs #522, #1696.

armaxri:
Settings/store foundation of the terminal syntax-highlighting epic (#1696), building on the merged engine foundation (#1697). Keeps the scope tight to what the rest of the chain (#1700/#1701/#1703/#1704) depends on.

## What this adds

- **Types** — `AppSettings.syntaxHighlighting?: SyntaxHighlightingConfig` (`src/types/connection.ts`) and `TerminalOptions.syntaxHighlighting?: ConnectionHighlightingConfig` (`src/types/terminal.ts`, where `TerminalOptions` actually lives). Both optional; absent config resolves to the built-in defaults, so older settings files stay forward-compatible.
- **Resolve helpers** (`src/services/syntaxHighlightingConfig.ts`, new file — no collision surface):
  - `defaultHighlightingConfig()` / `defaultBuiltinRuleFlags()` — shipped defaults: P0/P1 built-in rules on, P2/P3 off, no custom rules.
  - `resolveHighlightingConfig(global, perConnection)` — folds the global config and an optional per-connection override (`global` / `always-on` / `always-off`) into a `ResolvedHighlightingConfig`, appending per-connection `additionalRules` after the global custom rules.
  - `resolveActiveRules(config)` — the concrete enabled-rule list the engine compiles (builtin flags with shipped-default fallback + enabled custom rules), returned as copies so callers can recolor without mutating the shared built-in definitions.
- **Store state** (`src/store/appStore.ts`) — a runtime-only `sessionHighlighting: Record<sessionId, boolean>` map + `setSessionHighlighting(id, enabled|undefined)` action for the future status-bar quick toggle (child #1704). Never persisted; passing `undefined` clears an override. The global config lives in `settings.syntaxHighlighting` and persists via the existing `updateSettings`/`save_settings` path.
- **Backend persistence** (`src-tauri/src/connection/settings.rs`) — an additive `syntax_highlighting: Option<serde_json::Value>` field on the Rust `AppSettings` struct so the global config round-trips verbatim through settings save/restore. Stored opaquely (the config shape is owned by the frontend), following the existing `grammar: serde_json::Value` passthrough pattern. Without it the strict struct would silently drop the field on the save→load round-trip.

## Product default

The master switch ships **off** (opt-in) — the built-in rules keep their shipped P0/P1-on / P2/P3-off flags so enabling the switch immediately activates the sensible set. The issue left the default-on-vs-off choice open; flagging it here in case a default-on is preferred.

## Tests

- `src/services/syntaxHighlightingConfig.test.ts` — resolve helpers: global/always-on/always-off overrides, default fallback, additional-rule ordering, builtin-flag resolution, copy isolation.
- `src/store/appStore.syntaxHighlighting.test.ts` — per-session toggle set/clear + global-config round-trip through `updateSettings`.
- Rust: `syntax_highlighting_round_trip_preserves_opaque_config` + `syntax_highlighting_none_omitted_from_json` in `settings.rs`.

All frontend store tests (477), the resolve-helper suite, `cargo test connection::settings`, `tsc`, ESLint, Prettier, and `cargo clippy` pass locally.

Closes #1698. Part of #1696.

### DIFF
--- a/src-tauri/src/connection/settings.rs
+++ b/src-tauri/src/connection/settings.rs
@@ -223,6 +223,15 @@ pub struct AppSettings {
     /// `#[serde(default)]` keeps older settings files forward-compatible.
     #[serde(default)]
     pub shell_integration: ShellIntegrationSettings,
+    /// Terminal output syntax-highlighting configuration (epic #1696).
+    ///
+    /// The full config shape is owned by the frontend
+    /// (`src/types/syntaxHighlighting.ts` / `SyntaxHighlightingConfig`); the
+    /// backend only persists it verbatim, so it is stored as an opaque JSON
+    /// value here rather than mirrored as a typed struct. Absent → the frontend
+    /// resolves the built-in defaults.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub syntax_highlighting: Option<serde_json::Value>,
 }
 
 impl Default for AppSettings {
@@ -264,6 +273,7 @@ impl Default for AppSettings {
             updates: UpdateSettings::default(),
             serial_port_scan_prefixes: None,
             shell_integration: ShellIntegrationSettings::default(),
+            syntax_highlighting: None,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


armaxri:
Filed follow-up #1735 for out-of-scope work discovered here: several frontend `AppSettings` fields (`lineHeight`, `confirmCloseLiveSession`) are missing from the Rust `AppSettings` struct and are silently dropped on save. Not addressed in this PR to keep scope tight.

Issue #1765, opened 2026-07-19
feat(rdp): clipboard file transfer over CLIPRDR


armaxri:
Follow-up to #1757 (drive redirection PR) and #1756 (text clipboard). Deferred: #1756 landed text-only CLIPRDR and left file transfer as a follow-up; #1757 landed drive redirection (RDPDR), which is the transport CLIPRDR file transfer builds on.

## Context

The sidecar's `SidecarClipboardBackend` (`rdp-sidecar/src/clipboard.rs`) currently handles CF_UNICODETEXT / CF_TEXT only; its `on_file_contents_request` / `on_file_contents_response` callbacks are no-ops and it advertises no file formats. CLIPRDR file transfer (copy-paste of files between local and remote) uses the `CB_FILECONTENTS_REQUEST` / `CB_FILECONTENTS_RESPONSE` PDUs and the `FileGroupDescriptorW` format, and shares the same file-serving machinery as RDPDR drive redirection (now available after #1757).

## Scope

- [ ] Advertise the file-list clipboard format and negotiate the file-transfer CLIPRDR capability flag.
- [ ] Implement `on_file_contents_request` (serve local file bytes/size to the remote) and `on_file_contents_response` (receive remote file bytes) via the deferred-event pattern already used for text (`ClipboardEvent`).
- [ ] Bridge file clipboard data over the sidecar IPC and into the host clipboard / a chosen local directory.
- [ ] Honor view-only (never push local files to the remote in view-only sessions).
- [ ] Verify against a real RDP host (per-PR CI does not exercise the sidecar or a live server).

## Done when

A file copied on either side can be pasted on the other.

armaxri:
Landed the receiving direction in PR #1777 (merged): when opted in, files the remote copies to its clipboard are downloaded into the sandboxed shared folder (CLIPRDR `FileGroupDescriptorW` + File Contents PDUs, reusing #1757's sandbox). The remaining scope is now tracked as dedicated follow-ups, so closing this in favor of them:

- #1778 — serve local files to the remote (local→remote paste)
- #1779 — bridge the host OS clipboard's native file list (CF_HDROP / NSFilenames / text/uri-list) + IPC, on-demand paste
- #1780 — preserve directory structure + stream files larger than the 32 MiB cap

Issue #1757, opened 2026-07-19
feat(rdp): drive redirection and audio in the sidecar


armaxri:
Follow-up to #1747 (PR #1753).

**Scope**
- [ ] RDPDR drive redirection (share a local folder to the session).
- [ ] Audio playback redirection.
- [ ] Extend the IPC protocol and connection schema as needed.

**Done when** a redirected drive appears in the remote session and remote audio plays locally.

Issue #1756, opened 2026-07-19
feat(rdp): CLIPRDR clipboard bridging over the sidecar IPC


armaxri:
Follow-up to #1747 (PR #1753).

Clipboard is stubbed: `SidecarMessage::Clipboard`/`HostMessage::SetClipboard` exist in the protocol but the sidecar neither reads nor writes CLIPRDR, and `supports_clipboard` is `false`.

**Scope**
- [ ] Wire IronRDP CLIPRDR in the sidecar (text both ways).
- [ ] Emit `SidecarMessage::Clipboard` on server copy; apply `HostMessage::SetClipboard` on local copy.
- [ ] Flip `supports_clipboard` to `true`.

**Done when** text copied on either side is available on the other, honoring view-only.

PR #1777, opened 2026-07-19
feat(rdp): receive cliprdr clipboard files into the shared folder (Part of #1765)

armaxri:
## Summary

Extends the RDP sidecar's CLIPRDR bridge (#1756 landed text) to **receive files** the remote copies to its clipboard, downloading them into the sandboxed shared folder from #1757. Opt-in, off by default.

This is **`Part of #1765`** — it delivers the receiving direction; serving local files to the remote and bridging the host OS clipboard's native file list are deferred to follow-ups (see below). The issue stays open.

## What it does

- Advertises the CLIPRDR file-transfer capability (`CB_STREAM_FILECLIP_ENABLED`) + long format names so the server offers `FileGroupDescriptorW` when the remote copies files.
- On a remote file copy: detects the file-list format, fetches the list, and downloads each file's bytes (size → range, whole-file) via the same deferred-`ClipboardEvent` pattern text uses (the backend can't re-enter the borrowed `Cliprdr` channel, so the driver performs `initiate_paste` / `request_file_contents`).
- Writes received files into the **one opted-in shared folder** — the same folder drive redirection (#1757) already exposes read-write, so this grants the remote no new local access.

## Security / sandboxing

- Off by default; opt-in via the new **"Receive Clipboard Files"** setting, which requires drive redirection (reuses its share root as the destination).
- Every destination path goes through the shared sandbox choke point (extracted from #1757's drive resolver): rejects `..` traversal, drive letters/colons, NUL, and symlink escapes.
- Windows reserved device names (`CON`, `NUL`, …) are rejected; colliding basenames are deduplicated (`file (1).txt`); files over 32 MiB are skipped (bounds sidecar memory — whole-file range).
- The bridge only ever **receives**, so a view-only session (which must never push local data to the remote) is unaffected.

## Verification

- New unit tests cover: file-list format detection, capability gating, remote-copy → paste initiation, size-known and size-unknown download sequences, empty/oversized/directory/device-name handling, dedup, error-skips-to-next, and stream-id mismatch. Sandbox + config accessor/schema tests added.
- `rdp-sidecar` (excluded crate): `cargo test` 78 passing, `cargo clippy` + `cargo fmt` clean.
- Workspace `./scripts/check.sh` green; `pnpm markdownlint` clean.
- **Live end-to-end** (real RDP host) is not exercised by per-PR CI and needs manual verification: enable drive redirection + "Receive Clipboard Files", copy a file on the remote desktop, confirm it appears in the shared folder.

## Deferred (follow-ups, Part of #1765)

- Serving local files to the remote (local→remote paste).
- Host OS-clipboard-native file integration (CF_HDROP / NSFilenames / text/uri-list) + IPC, so arbitrary host-clipboard files transfer both ways and download is tied to a real local paste gesture.
- Preserve copied directory structure (currently flattened) + chunked streaming for files over the size cap.

### DIFF
--- a/core/src/backends/rdp_sidecar/config.rs
+++ b/core/src/backends/rdp_sidecar/config.rs
@@ -103,6 +103,13 @@ pub struct RdpConfig {
     /// (#1764). Audible playback is currently macOS/Windows only — the Linux
     /// build omits the audio backend (see the #1764 Linux follow-up).
     pub audio_redirection: bool,
+    /// Opt in to CLIPRDR file transfer (#1765): when the remote copies files to
+    /// its clipboard, download them into the shared folder ([`Self::shared_folder_path`]).
+    /// Requires [`Self::drive_redirection`] and a valid shared folder — that
+    /// folder is already remote-writable via drive redirection, so receiving
+    /// clipboard files into it grants the remote no new local access. Off by
+    /// default.
+    pub clipboard_file_transfer: bool,
 }
 
 impl Default for RdpConfig {
@@ -123,6 +130,7 @@ impl Default for RdpConfig {
             shared_folder_path: String::new(),
             drive_name: String::new(),
             audio_redirection: false,
+            clipboard_file_transfer: false,
         }
     }
 }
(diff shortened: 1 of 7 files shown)


armaxri:
Follow-ups filed (Part of #1765): #1778 (serve local→remote), #1779 (host OS-clipboard native file list + IPC), #1780 (preserve directory structure + stream large files).

PR #1811, opened 2026-07-20
feat(rdp): multi-format rdpsnd negotiation via vendored ironrdp-rdpsnd (Closes #1773)

### LINKED ISSUE #1773: feat(rdp): multi-format rdpsnd negotiation (more PCM rates, ADPCM/Opus)
Follow-up to #1764 (PR to be linked).

#1764 advertises a **single** PCM audio format (44.1 kHz / 16-bit / stereo) over rdpsnd. This is deliberate: IronRDP 0.9's `RdpsndClientHandler::wave(format_no, ...)` reports an index into the *intersection* of the client and server format lists, which IronRDP builds from a `HashSet` (non-deterministic order) and does not expose to the handler. With more than one advertised format the handler cannot reliably map `format_no` back to a concrete format (sample rate / channels), so a single format keeps `format_no` unambiguous (always 0).

If the server does not offer exactly 44.1 kHz/16-bit/stereo PCM, no audio plays.

## Scope

- [ ] Advertise additional common PCM formats (e.g. 22.05 kHz, mono) to widen server compatibility, solving the `format_no`→format mapping. Likely needs an upstream IronRDP change so the handler is told the concrete `AudioFormat` for each `wave` (or the exact client format list that was sent) — investigate and, if needed, file/track upstream.
- [ ] Optionally decode compressed formats the server prefers (ADPCM, Opus) rather than forcing PCM.

## Done when

The sidecar negotiates audio with servers that do not offer 44.1 kHz/16-bit/stereo PCM, and plays the correct sample rate/channels for whichever format the server selects.

armaxri:
Closes #1773

## Summary

Enables sound **multi-format `rdpsnd` (RDP audio) negotiation**. The sidecar now advertises a table of common PCM formats (48 / 44.1 / 22.05 / 11.025 kHz, mono and stereo, 16-bit) and plays each `wave` buffer at exactly the channel count and sample rate the server negotiated — instead of forcing everything through one fixed 44.1 kHz stereo format.

## Why this needed a vendored crate

Issue #1773 was blocked on an upstream `ironrdp-rdpsnd` API limitation (documented by a prior investigation on the issue): upstream 0.9.0 hands the client `wave` callback only a `format_no` **index** into the Client Audio Formats list, and it builds that list from a `HashSet` intersection (non-deterministic per-process order) then **discards** it. With more than one advertised format, the handler cannot map a played buffer back to its negotiated rate — so advertising multiple rates risked "chipmunk" audio (e.g. playing 22.05 kHz data at 44.1 kHz). Advertising a single format was the only sound option.

Per the maintainer's decision, this **vendors and patches `ironrdp-rdpsnd`** rather than waiting on upstream — the same pattern as the existing `vendor/vnc-rs` fork.

## What changed vs upstream (the vendored fork)

`rdp-sidecar/vendor/ironrdp-rdpsnd/` is a fork of upstream 0.9.0, patched into the workspace-excluded sidecar graph via `[patch.crates-io]` (the meta `ironrdp` crate depends on `ironrdp-rdpsnd = "0.9"` and re-exports it, so the patch redirects that re-export). The **only functional change** (in `src/client.rs`):

- `Rdpsnd` now remembers the exact, order-preserving `Vec<AudioFormat>` it sent in `client_formats()` (`negotiated_formats`).
- The trait method becomes `RdpsndClientHandler::wave(&mut self, format: &AudioFormat, ts, data)` — passing the **concrete negotiated format** instead of an unresolvable index. `Wave2 { format_no }` is resolved against the remembered list; an out-of-range index is dropped, not guessed.

`pdu/mod.rs`, `server.rs` and `lib.rs` are byte-for-byte upstream. Sibling `ironrdp-*` deps stay registry versions, so the fork adds **no new locked crate** (verified: the only `Cargo.lock` change is the patched crate losing its registry source/checksum; `visibility` stays out of the lock). This is exactly the change intended to be offered upstream to Devolutions/IronRDP. See `vendor/ironrdp-rdpsnd/README.md`.

## Sidecar side (`rdp-sidecar/src/audio.rs`)

- Advertises the PCM format table; decodes and plays each buffer at its own negotiated `(channels, sample_rate)` — rodio resamples to the device rate.
- The jitter/latency buffer is now **format-aware** (durations computed from each buffer's own rate/channels), so a mid-session format change cannot introduce a pitch/rate error.
- Non-PCM and non-16-bit buffers are rejected rather than mis-played.

## Soundness guard (no chipmunk-audio regression)

A dedicated test asserts the load-bearing property: the **same bytes** decoded under a 22.05 kHz mono format report 22 050 Hz / 1 channel, and under a 48 kHz stereo format report 48 000 Hz / 2 channels — proving the playback rate comes from the negotiated format, never a fixed assumption. Plus tests for the format table's PCM derivations, format rejection, and the existing jitter-buffer invariants (updated to be format-aware). All 12 audio tests pass.

## Cross-platform

No new dependencies; `rodio` and its target gating are unchanged. `cargo tree -i ironrdp-rdpsnd` shows only the meta crate as dependent. The vendored fork lives entirely in the workspace-excluded sidecar graph, so it cannot affect the desktop app's dependency resolution.

## Deferred

ADPCM / Opus decode is higher-scope (needs a mature, cleanly-integrable, correctness-verifiable decoder crate) and is filed as a follow-up. Multi-PCM-rate negotiation is the safe, high-value core and ships here.

## Testing

- `cargo build --manifest-path rdp-sidecar/Cargo.toml` — clean.
- `cargo test --manifest-path rdp-sidecar/Cargo.toml audio::` — 12/12 pass.
- `cargo clippy --manifest-path rdp-sidecar/Cargo.toml --all-targets` — clean.
- `cargo fmt --manifest-path rdp-sidecar/Cargo.toml --check` — clean.

Note: the sidecar is built only by the bundle/release CI, not per-PR CI, so the above were run locally.

### DIFF
--- a/docs/changes/dev0/feature/1773-multiformat-rdpsnd.md
+++ b/docs/changes/dev0/feature/1773-multiformat-rdpsnd.md
@@ -1,0 +1,8 @@
+### Added
+
+- RDP audio (`rdpsnd`) now negotiates **multiple PCM formats** instead of a
+  single fixed one. The sidecar advertises a table of common sample rates
+  (48/44.1/22.05/11.025 kHz) in both mono and stereo (16-bit PCM) and plays each
+  audio buffer at exactly the rate and channel count the server selected, so
+  audio from servers that stream at a non-CD rate is no longer forced through a
+  single 44.1 kHz stereo assumption (#1773).
--- a/rdp-sidecar/Cargo.lock
+++ b/rdp-sidecar/Cargo.lock
@@ -2139,8 +2139,6 @@ dependencies = [
 [[package]]
 name = "ironrdp-rdpsnd"
 version = "0.9.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1d1a4f7092ae8fdc0a2f61be8e100277dc6f4ad51b85455dc9e8157979993ea2"
 dependencies = [
  "bitflags 2.13.1",
  "ironrdp-core",
--- a/rdp-sidecar/Cargo.toml
+++ b/rdp-sidecar/Cargo.toml
@@ -171,3 +171,18 @@ wl-clipboard-rs = { version = "0.8", default-features = false, features = [
 # Temp directories back the drive-redirection filesystem tests (#1757), which
 # exercise the RDPDR backend against a real folder without a live RDP server.
 tempfile = "3"
+
+# Vendored fork of ironrdp-rdpsnd 0.9.0 (#1773), patched in the same way as
+# `vendor/vnc-rs` is patched into the main app (#1714). Upstream 0.9.0 hands the
+# `wave` callback only a `format_no` index into a non-deterministically-ordered,
+# then-discarded client-format list, so a handler advertising more than one audio
+# format cannot map a played buffer back to its negotiated rate (risking
+# "chipmunk" audio). The fork remembers the exact list it sent and passes the
+# concrete `AudioFormat` to `wave`, which is what makes multi-format rdpsnd
+# negotiation sound. `ironrdp` (the meta crate) depends on `ironrdp-rdpsnd = "0.9"`
+# and re-exports it as `ironrdp::rdpsnd`, so patching crates-io redirects that
+# re-export to the fork. This lives in the workspace-EXCLUDED sidecar graph with
+# its own Cargo.lock, so it never touches the desktop app's dependency graph.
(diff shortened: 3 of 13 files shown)


armaxri:
Follow-up filed for compressed audio (ADPCM / Opus): #1812

PR #1817, opened 2026-07-21
feat(rdp): rdpsnd compressed audio formats — ms-adpcm / ima-dvi adpcm (Closes #1812)

### LINKED ISSUE #1812: feat(rdp): rdpsnd compressed audio formats (ADPCM / Opus)
Follow-up to #1773 (PR #1811).

#1773 landed multi-format PCM negotiation for `rdpsnd` (48/44.1/22.05/11.025 kHz, mono/stereo, 16-bit) on top of a vendored, patched `ironrdp-rdpsnd` that now passes the concrete negotiated `AudioFormat` to the `wave` callback. Compressed formats were deliberately deferred as higher-scope.

## Scope

- [ ] Advertise and decode **MS-ADPCM** (`WAVE_FORMAT_ADPCM`, 0x0002) and/or **DVI/IMA ADPCM** (0x0011) — the codecs Windows `rdpsnd` servers commonly offer besides PCM. The negotiated `AudioFormat.data` carries the codec's `cbSize` extra bytes (coefficients / samples-per-block) needed to decode.
- [ ] Optionally advertise/decode **Opus** if a server offers it.
- [ ] Integrate a **mature, maintained decoder crate**, `default-features = false`, target-gated to macOS/Windows/Linux, and `cargo tree -i` clean off other platforms (same bar as the existing sidecar audio deps). Justify the crate choice; prefer a library over a hand-rolled decoder (repo policy).
- [ ] Verify correctness — a decode round-trip / known-vector test proving the codec plays at the right rate (extend the format-correctness guard from #1773 so the "chipmunk audio" class stays covered).
- [ ] Wire the new formats into `advertised_formats()` and `decode_wave()` in `rdp-sidecar/src/audio.rs`; the vendored crate already exposes the concrete format, so no further ironrdp-rdpsnd change is expected.

## Done when

A server that offers ADPCM (and/or Opus) has that format negotiated and played back correctly at its true sample rate, verified by a test, with the decoder dependency cleanly target-gated and adding no cross-platform build breakage.

armaxri:
Follow-up to #1773 (PR #1811). Adds **compressed** `rdpsnd` audio playback on top of the multi-rate PCM path.

Closes #1812.

## What ships

- **MS-ADPCM** (`WAVE_FORMAT_ADPCM`, 0x0002) and **IMA/DVI ADPCM** (`WAVE_FORMAT_DVI_ADPCM`, 0x0011) are now advertised, negotiated, and decoded to 16-bit PCM before the existing jitter buffer + rodio sink.
- Each `wave` payload is decoded at the format's **own** sample rate and channel count, so the #1773 "chipmunk audio" guard (play at the negotiated rate, never a fixed assumption) holds for compressed formats too.

**Opus is deferred**, not shipped — its maintained decoders link the native `libopus` C library, a cross-platform CI cost. Filed as a dedicated follow-up (linked below). Opus was optional in the issue ("and/or"), and the "Done when" criterion (a server offering ADPCM has it negotiated and played correctly, verified by a test) is met by this PR.

## Decoder choice

Decoding is delegated to **Symphonia** (`symphonia-codec-adpcm` + `symphonia-core`) rather than hand-rolled, per the repo's "prefer maintained libraries" policy. Symphonia is the pure-Rust, widely-used multimedia library; its ADPCM codec covers both formats the issue names.

- Added only to the **workspace-excluded** sidecar graph (its own `Cargo.lock`), target-gated to macOS/Windows/Linux alongside `rodio`, `default-features = false`.
- **No native/C build dependency** and `cargo tree` clean: the two crates pull only pure-Rust deps already present in the graph; the lock gains exactly `symphonia-core` and `symphonia-codec-adpcm`. Zero impact on the desktop app's dependency graph.

## Why the vendored `ironrdp-rdpsnd` fork needed a change

The issue anticipated no further fork change, but the negotiation is an exact-equality `HashSet` intersection over the whole `AudioFormat`. A compressed format's block layout (`nBlockAlign`, `wSamplesPerBlock`, and MS-ADPCM coefficient `data`) is **server-chosen and unpredictable**, so a fixed client advertisement can never match it. The minimal fix:

- Adds `RdpsndClientHandler::accepts_format` (default = the prior exact-match behaviour).
- Rewrites `client_formats()` to iterate the **server's** formats in order and echo back the accepted ones **verbatim** — so `wave` still receives concrete, decodable parameters. Also makes the sent list deterministic where the `HashSet` was not.

`RdpAudioBackend::accepts_format` only accepts an ADPCM format the decoder can actually handle (supported codec, mono/stereo, non-zero rate, resolvable block size), so we never advertise a format we would then drop.

## Correctness guard (per format)

Decode-correctness tests assert exact PCM output against closed-form ADPCM vectors — blocks whose zero-value nibbles hold the verbatim preamble sample(s), independent of the adaptation math — at the negotiated rate, for both codecs. Plus multi-block decode, the frames-per-block formulas (mono + stereo, both codecs), and rejection of unsupported/unusable formats (dropped, never mis-decoded).

## Verification

The sidecar is workspace-excluded, so per-PR CI (`cargo clippy --workspace`) does not cover it. Validated **locally in `rdp-sidecar/`**:

- `cargo test` — 169 pass (7 new ADPCM tests)
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean

## Test plan

- [x] Unit: MS-ADPCM mono decodes to expected PCM at 22.05 kHz / 1 ch
- [x] Unit: IMA/DVI ADPCM mono decodes to expected PCM at 11.025 kHz / 1 ch
- [x] Unit: multi-block decode; frames-per-block formulas; extra-data `wSamplesPerBlock` path
- [x] Unit: unsupported/unusable formats rejected, not mis-decoded
- [ ] Manual: connect to a Windows RDP server offering ADPCM and confirm audible, correctly-pitched playback (requires a live server; no ADPCM-offering server available in CI)

### DIFF
--- a/docs/changes/dev0/feature/1812-rdpsnd-compressed-audio.md
+++ b/docs/changes/dev0/feature/1812-rdpsnd-compressed-audio.md
@@ -1,0 +1,11 @@
+### Added
+
+- RDP audio redirection now plays **compressed** `rdpsnd` streams in addition to
+  PCM: an RDP server that offers **MS-ADPCM** (`WAVE_FORMAT_ADPCM`) or **IMA/DVI
+  ADPCM** (`WAVE_FORMAT_DVI_ADPCM`) has that format negotiated and decoded to
+  16-bit PCM before playback, at the format's own sample rate and channel count
+  (so the multi-rate "chipmunk audio" guard from #1773 still holds). Decoding uses
+  the pure-Rust Symphonia ADPCM codec, target-gated to macOS/Windows/Linux like
+  the rest of the sidecar audio path and adding no native build dependency. Opus
+  (`WAVE_FORMAT_OPUS`) remains unsupported for now — its decoders link the native
+  `libopus` C library — and is tracked as a follow-up (#1812).
--- a/rdp-sidecar/Cargo.lock
+++ b/rdp-sidecar/Cargo.lock
@@ -4168,6 +4168,29 @@ dependencies = [
 ]
 
 [[package]]
+name = "symphonia-codec-adpcm"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2dddc50e2bbea4cfe027441eece77c46b9f319748605ab8f3443350129ddd07f"
+dependencies = [
+ "log",
+ "symphonia-core",
+]
+
+[[package]]
+name = "symphonia-core"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ea00cc4f79b7f6bb7ff87eddc065a1066f3a43fe1875979056672c9ef948c2af"
+dependencies = [
+ "arrayvec",
+ "bitflags 1.3.2",
+ "bytemuck",
+ "lazy_static",
+ "log",
+]
+
+[[package]]
 name = "syn"
 version = "1.0.109"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4328,6 +4351,8 @@ dependencies = [
  "percent-encoding",
  "rodio",
  "sha2 0.11.0",
+ "symphonia-codec-adpcm",
+ "symphonia-core",
  "tempfile",
  "termihub-core",
  "tokio",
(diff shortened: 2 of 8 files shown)


armaxri:
Opus decode (deferred from this PR) is tracked as follow-up #1818.

PR #1840, opened 2026-07-23
feat(config): per-connection environment variables for local shells

### LINKED ISSUE #1825: feat(config): per-connection environment variables for spawned terminals
### Description
Feature request: allow defining per-connection environment variables so that a terminal spawned for that connection starts with those variables set in its environment.

### Scope
- [ ] Add a per-connection env-vars editor (key/value pairs) in the connection settings.
- [ ] Persist the env vars with the connection config.
- [ ] Inject the configured env vars into the spawned terminal's environment for the relevant connection types (local shell, and where applicable serial/agent/SSH).
- [ ] Document precedence/override behavior with the inherited environment.

### Done when
A user can define environment variables on a connection and terminals spawned from it start with those variables set.

armaxri:
## Summary

Implements per-connection environment variables so a local shell connection spawns its terminal with user-defined env vars set. Requested in the issue: "Setting env variables for a connection would be super, so that a terminal spawns with specific env vars."

Closes #1825

## What changed

- **Schema (`core/src/backends/local_shell.rs`)** — adds an **Environment** group with an `envVars` field of type `KeyValueList`. Because connection forms are schema-driven, `DynamicForm` renders the key/value editor automatically (the same widget Docker already uses for container env vars) — no frontend changes were needed.
- **Parsing** — `connect()` parses the `envVars` list into `ShellConfig.env`.
- **Spawn threading** — `build_shell_command` already copies `config.env` into `ShellCommand.env`, and the PTY spawner applies each var on top of the inherited environment via `CommandBuilder::env`. **Precedence:** a connection env var overrides an inherited value of the same name; all other inherited vars are preserved.
- **Expansion** — `ShellConfig::expand` now expands `${VAR}` in env *values* (e.g. `${HOME}/bin`), matching how Docker expands its env vars. Keys are left verbatim.

## Connection types covered

- **Local shell** — covered (this PR). Works on the desktop and agent paths since both use the shared `core` backend.
- **Docker** — already had container env vars; unchanged.
- **SSH / WSL** — scoped out to a follow-up. Client-side env vars do not transfer to an SSH server unless it opts in via `AcceptEnv`/`SetEnv`, and WSL env crossing the Win32/WSL boundary needs `WSLENV`. Shipping the local-shell field for these without that handling would silently do nothing, so they get correct, dedicated handling in a follow-up.
- **Serial / Telnet / FTP** — not applicable (no locally-spawned shell process).

## Testing

- `cargo test -p termihub-core --all-features` — 941 pass, including new tests:
  - `parse_env_vars_*` (parsing + skipping incomplete entries)
  - `schema_has_env_vars_key_value_field`
  - `connect_applies_env_vars_to_spawned_command` — end-to-end via mock spawner, asserts the connection var reaches the spawned command's environment alongside `TERM`
  - `shell_config_expand_expands_env_values`
- `./scripts/check.sh` — pass
- `pnpm test` — 3833 pass

### DIFF
--- a/core/src/backends/local_shell.rs
+++ b/core/src/backends/local_shell.rs
@@ -5,6 +5,7 @@
 //! calls `portable_pty::native_pty_system()`; tests inject `MockLocalShellSpawner`
 //! which returns in-memory pipes and never forks a real process.
 
+use std::collections::HashMap;
 use std::io::{Read, Write};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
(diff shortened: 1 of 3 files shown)


armaxri:
Follow-up filed for SSH/WSL per-connection env vars: #1841

Issue #1899, opened 2026-07-25
[Epic] Multi-window support (detachable / movable tabs across windows)


armaxri:
> **Tracking epic — do not assign this issue to a dev slot.** It has no code of its own. Implement the child issues below **in dependency order**; each child (except this epic) is separately labelled `Ready2Implement`.

## Goal

Add **multi-window support** to termiHub: let a user host **multiple native windows** and **move a live session tab between them without tearing down the backend session** (its PTY / SSH channel / serial port / RDP connection keeps running). Three user-facing capabilities:

1. **Tear a tab out** into a brand-new window ("Move to New Window").
2. **Create** additional windows (empty, or seeded).
3. **Move tabs between** existing windows ("Move to Window ▸ N"), preserving scrollback to the extent the backend replay buffer allows.

Source of truth: the design-first concept **`docs/concepts/backlog/multi-window.html`** (authored from #1806 — senior-engineer technical concept + UX mockups). **Read it before implementing** — the re-parenting model, the state-ownership/event-retargeting table, the per-session-type feasibility matrix, the per-platform constraints, the states/sequences diagrams, the limitations, and the six inline mockups are all there. The concept is authoritative; when it and the code disagree, fix the code.

## Why this is feasible (and where it is hard)

The architecture is **unusually well-suited** in one dimension: session *state* already lives in the Rust backend keyed by `session_id`, completely independent of any window (`SessionManager`, `GraphicalSessionManager`), and output is a **global broadcast** (`AppHandle::emit("terminal-output", …)`) already delivered to every webview. The backend already supports the exact primitive multi-window needs — **detach a view and re-attach it later with scrollback replay** (1 MiB `RingBuffer` + persistent/agent replay). So re-parenting a tab is a **pure view operation**, not a teardown.

The **hard limits cluster on gesture and coordination**, not session survival:

- Separate native windows are **separate JS contexts** — no shared DOM, no shared Zustand store.
- **`@dnd-kit` is single-document** and cannot follow the cursor across a native window boundary, and Tauri exposes no webview→webview drag → tear-out/move ship as **commands, not live cross-window drag**.
- Graphical (RDP/VNC) framebuffers are canvas memory in the source window → command-only move with a repaint flash.
- Persistence has **no window dimension** today.
- Window-close policy is **new** (detach-vs-terminate; per-OS quit rules).

## Why sequenced (the collision risk)

[...]

PR #1963, opened 2026-07-25
feat(sftp): scope transfer-queue ownership to the owning window on tab move

### LINKED ISSUE #1951: feat(multi-window): scope transfer-queue ownership to the owning window on tab move
Deferred item from the multi-window epic #1899 (concept now implemented). The SFTP/FTP `transferQueue` is app-scoped (`src/store/appStore.ts` ~1005) with no window-ownership handling. When a transfer-bearing tab is moved to another window (moveTabToWindow), its queued/active transfers should follow the tab / be attributed to the owning window rather than staying globally app-scoped.

## Scope
- Decide + implement how in-flight and queued transfers are attributed when a tab re-parents across windows (follow the tab, or remain where started with clear UI). Reuse #1900's `session_id → owning_window` ownership map.
- Cover the Open-Connections / transfer UI so a transfer's owning window is correct after a move.

## Done when
- Moving an SFTP/FTP tab to another window keeps its transfers correctly attributed (no orphaned/duplicated queue entries), with tests.

Deferred from epic #1899; the concept explicitly left transfer-queue ownership on move as an open question.

armaxri:
## Summary

Deferred item from the multi-window epic (#1899): the SFTP/FTP Transfer Queue was app-scoped, with no window-ownership handling. Because each native window is a separate JS context with its own Zustand store and `transfer-progress` events are `app.emit`-broadcast to every window, a transfer-bearing tab's queued/active transfers stayed orphaned in the source window when the tab moved to another window.

This makes **transfers follow the tab**:

- On `moveTabToWindow` / `moveWindowSessionsToWindow`, the tab's Transfer Queue rows (rows whose `sessionId` is the tab's own session — FTP `file-browser` tabs — or an SFTP sidebar session bound to the tab via `sftpSessions[…].owningTabId`) are carried in the opaque frontend-owned hand-off envelope (`HandoffTab.transfers`). The backend just ferries `tab`, so no Rust change.
- The **destination** window seeds the carried rows into its own `transferQueue` on hydrate (existing rows win, so a broadcast event that already advanced a row is not clobbered) and un-releases those sessions so it resumes folding their live progress.
- The **source** window drops the moved rows from both the persistent `transferQueue` and the transient `transfers` map, and marks the moved session ids `releasedTransferSessions` so ongoing broadcast `transfer-progress` events do not re-adopt the moved rows (no orphaned/duplicated entries).

The released-set is empty by default, so single-window behavior and all existing tests are unchanged.

## Scope note / follow-up

This scopes ownership **on move** (the issue's "Done when"). It does not change the pre-existing app-wide broadcast behavior whereby a window that never participated in a move still folds every broadcast transfer; fully scoping the display to the owning window even without a move needs the backend `session → window` map surfaced to the fold and is filed as a follow-up.

## Test plan

- New `src/store/appStore.transferHandoff.test.ts`: moving a transfer-bearing tab carries its rows to the destination, removes them from the source, releases the session so a later broadcast event does not re-create the row, and un-releases on hydrate/seed. Whole-window move carries per-tab transfers.
- Existing transfer/multi-window/sftp store tests still green.

Concept updated (`docs/concepts/implemented/multi-window.html`) to mark decision 6 resolved.

Closes #1951

### DIFF
--- a/docs/changes/dev2/feat/1951-transfer-queue-window-ownership.md
+++ b/docs/changes/dev2/feat/1951-transfer-queue-window-ownership.md
@@ -1,0 +1,8 @@
+### Changed
+
+- Multi-window: an SFTP/FTP tab's file transfers now **follow the tab** when it
+  is moved to another window. The tab's Transfer Queue rows are carried with the
+  hand-off and seeded into the destination window's queue, while the source
+  window drops them and stops adopting the moved session's later transfers — so
+  moving a transfer-bearing tab no longer leaves orphaned or duplicated queue
+  entries in the original window (#1951).
--- a/docs/concepts/implemented/multi-window.html
+++ b/docs/concepts/implemented/multi-window.html
@@ -1984,7 +1984,8 @@ sequenceDiagram
           </li>
           <li>
             <strong>Transfer-queue ownership</strong> when an SFTP/FTP tab with in-flight transfers
-            moves windows.
+            moves windows. <em>Resolved (#1951): transfers follow the tab</em> — carried in the
+            hand-off and seeded in the destination; see the deferred-items list below.
           </li>
           <li>
             <strong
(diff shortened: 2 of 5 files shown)


armaxri:
Filed follow-up #1964 for fully scoping transfer display to the owning window even without a tab move (the app-wide broadcast case this PR intentionally leaves out of scope).

Issue #2018, opened 2026-07-26
Plugin host-mediated capability bridge (enforce network/filesystem permissions at runtime over the ABI)


armaxri:
Follow-up to #2001 (PR #2016).

#2001 added the permission-enforcement primitives: a `PermissionSet` per session, `require_permission`, and a path-scoped `FilesystemScope::check`. But a native plugin backend runs **in the host process** and does its own I/O, so the host cannot currently intercept a plugin's own network/filesystem syscalls. The primitives are enforced today only at the surfaces the host mediates (load-time consistency, and any host-resolved path). Concept §13 ("The host can deny network access even if requested") implies a host-provided capability channel.

## Scope
- [ ] Extend the plugin ABI (`termihub-plugin-api`) so a plugin performs network/filesystem access **through host-provided callbacks** rather than directly.
- [ ] Pass the granted `PermissionSet` (and filesystem scope) to the plugin at `create_backend`.
- [ ] Route those host callbacks through `PermissionSet::require` / `FilesystemScope::check` so `network`/`filesystem` are enforced at runtime, not just declared.
- [ ] Tests: a plugin denied `network` cannot open a connection via the host bridge; a filesystem read outside the declared scope is rejected end-to-end.

## Done when
A plugin's network and filesystem access is actually mediated and can be denied by the host at runtime, exercising the `PermissionSet`/`FilesystemScope` guards added in #2001.

PR #2027, opened 2026-07-26
feat(plugin): mediate filesystem writes, stat and list_dir through the host bridge (Closes #2024)

### LINKED ISSUE #2024: Extend the plugin capability bridge: mediated filesystem writes and a persistent mediated socket
Follow-up to #2018 (PR #2023).

#2018 landed the host-mediated capability bridge (`PluginHostBridge`) that enforces `network`/`filesystem` permissions at runtime for the operations a plugin routes through it. To keep that PR a reviewable slice, the mediated surface was deliberately minimal:

- **Network:** `open_connection` mediates *connection establishment* (host enforces `network`, then hands back a `HostTcpStream`). Good enough for a plugin that dials out, but the host does not inspect traffic and there is no host-side connect timeout/policy knob.
- **Filesystem:** only a one-shot `read_file` is mediated. There is no mediated **write**, append, create, list, or stat — a plugin needing to write within its declared scope still has no enforced path.

## Scope
- [ ] Add mediated filesystem write/append/create through the bridge, routed through `FilesystemScope::check` (writes confined to declared paths).
- [ ] Consider mediated directory listing / metadata within scope.
- [ ] Optionally add host-side connection policy (timeout, per-session limits) to `open_connection`.
- [ ] Extend the `plugin_capability_bridge.rs` end-to-end tests to cover the new operations (allow + deny).
- [ ] Bump the plugin ABI version if the bridge vtable changes.

## Done when
A plugin can write files (within its declared scope) and the host rejects out-of-scope writes end-to-end, exercised over the real ABI boundary like the read path in #2018.

Note: closing the *direct-syscall bypass* (an in-process plugin calling `std::fs`/`std::net` directly, sidestepping the bridge) is explicitly **out of scope** here — it needs OS-level process isolation termiHub has no substrate for, and is already documented as future work in `core/src/plugin/security.rs`.

armaxri:
Follow-up to #2018 (PR #2023). Extends the deliberately-minimal host-mediated capability bridge (`PluginHostBridge`) with the write/inspect filesystem surface the read-only slice deferred.

## What changed

- **Mediated filesystem writes** — the bridge gains `write_file(path, data, mode)` with a `PluginWriteMode` of `Truncate` (create-or-overwrite), `Append`, or `CreateNew`. Every write resolves its path through `FilesystemScope::check` (via `PermissionSet::check_path`), so a plugin can only write inside its declared roots; out-of-scope and `..`-traversal paths are refused before any file is created or opened. Ergonomic plugin-side wrappers: `overwrite`, `append`, `write_new`.
- **Mediated directory inspection** — `stat(path)` returns FFI-safe `PluginFileMetadata` (exists / is_dir / len; an in-scope-but-absent path is reported as absent, not an error), and `list_dir(path)` returns the scoped directory's entry names. Both go through the same scope guard.
- **Host-side connection policy** — the mediated `open_connection` now applies a 30s host-side connect timeout (`connect_timeout` over resolved addresses), so a plugin's dial-out to a black-holed host can no longer hang a session.
- **ABI bump to 3** — the `PluginHostBridge` vtable grew three `#[repr(C)]` `extern "C"` callbacks, a layout change that breaks version-2 plugins, so `CURRENT_PLUGIN_API_VERSION` is bumped and the FFI-safety compile-time proof in `symbols.rs` covers the new shapes. The stable-ABI approach is preserved — no `dyn Trait`, `String`, or allocator crosses the boundary.

## Tests

- Host-side unit tests in `core/src/plugin/capabilities.rs` for write/append/create-new, stat, and list — each with allow **and** deny (out-of-scope, no-permission) paths.
- End-to-end tests in `core/tests/plugin_capability_bridge.rs` exercise the new operations over the **real `dlopen` boundary** via the fixture `cdylib` (new `writefile`/`appendfile`/`createfile`/`statpath`/`listdir` probes), asserting allow and deny for each.
- The host round-trip test's ABI-version assertions now reference the crate constant instead of a literal, so future bumps stay green.

## Out of scope

Closing the direct-syscall bypass (an in-process plugin calling `std::fs`/`std::net` directly) — that needs OS-level process isolation termiHub has no substrate for, and stays documented as future work in `core/src/plugin/security.rs`. See the follow-up below for per-session connection limits.

Closes #2024

### DIFF
--- a/core/src/plugin/capabilities.rs
+++ b/core/src/plugin/capabilities.rs
@@ -23,13 +23,6 @@ use std::net::{TcpStream, ToSocketAddrs};
 use std::path::Path;
 use std::time::Duration;
 
-/// Host-side connect timeout applied to every mediated `open_connection` (#2024).
-///
-/// A bare `TcpStream::connect` blocks indefinitely on a black-holed host; this
-/// per-connect ceiling is the host-side connection policy that keeps a plugin's
-/// dial-out from hanging a session forever.
-const CONNECT_TIMEOUT: Duration = Duration::from_secs(30);
-
 use termihub_plugin_api::capabilities::{
     PluginBridgeDestroyFn, PluginFileMetadata, PluginListDirFn, PluginOpenConnectionFn,
     PluginReadFileFn, PluginStatPathFn, PluginWriteFileFn, PluginWriteMode,
@@ -40,6 +33,13 @@ use termihub_plugin_api::{
 
 use super::security::{PermissionError, PermissionSet};
 use super::PluginPermission;
+
+/// Host-side connect timeout applied to every mediated `open_connection` (#2024).
+///
+/// A bare `TcpStream::connect` blocks indefinitely on a black-holed host; this
+/// per-connect ceiling is the host-side connection policy that keeps a plugin's
+/// dial-out from hanging a session forever.
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// Build the host capability bridge for a session granted `permissions`.
 ///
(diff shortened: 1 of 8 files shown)


armaxri:
Follow-up filed: #2028 — per-session connection limits and configurable connect timeout (the deferred half of #2024's host-side connection policy).

PR #2029, opened 2026-07-26
feat(plugin): per-session connection limits and configurable connect timeout on the bridge (Closes #2028)

### LINKED ISSUE #2028: Plugin capability bridge: per-session connection limits and policy
Follow-up to #2024 (PR #2027).

#2024 added a host-side **connect timeout** (30s) to the mediated `open_connection` on the `PluginHostBridge`. The other half of "host-side connection policy" from #2024's optional scope was left out to keep that PR reviewable: there is still no cap on how many concurrent mediated connections a single plugin session may open, nor a configurable timeout.

## Scope
- [ ] Track mediated `HostTcpStream`s per session and enforce a maximum concurrent-connection count (refuse with `PluginStatus::PermissionDenied`/a dedicated status once exceeded).
- [ ] Make the connect timeout configurable rather than the hardcoded `CONNECT_TIMEOUT` constant in `core/src/plugin/capabilities.rs`.
- [ ] Consider surfacing these limits through the manifest / permission declaration.

## Done when
A plugin session cannot exhaust host resources by opening unbounded connections through the bridge, and the connect timeout is configurable, both covered by tests.

armaxri:
Follow-up to #2024 (PR #2027). Adds the per-session **connection policy** that #2024 deliberately left out to keep that PR reviewable: a cap on how many concurrent mediated connections a plugin session may hold open, and a configurable connect timeout in place of the hardcoded 30s constant.

## What changed

- **Per-session connection ceiling** — the `PluginHostBridge` context now carries the session's `ConnectionPolicy` and a live connection counter. `open_connection` reserves a slot before dialing and refuses with `PluginStatus::PermissionDenied` once the session is at its ceiling (default **8**). The slot is released the instant the plugin drops the connection — via a host-side `StreamDropGuard` that rides on the mediated stream — or immediately if the connect itself fails, so a failed dial-out never leaks a slot.
- **Configurable connect timeout** — the previously hardcoded `CONNECT_TIMEOUT` is now `ConnectionPolicy::connect_timeout` (default **30s**), threaded into the mediated connect.
- **Manifest surface** — an optional `connectionPolicy` block (`maxConnections`, `connectTimeoutMs`) in `manifest.json` lets a plugin raise or lower either limit; any unset field falls back to the host default. `build_host_bridge_with_policy` is the new policy-aware constructor (`build_host_bridge` stays as a default-policy convenience).
- **No ABI bump** — `CURRENT_PLUGIN_API_VERSION` stays at **3**. The policy is enforced entirely host-side, and the new drop guard lives behind the mediated stream's opaque `state` pointer, so no vtable or `#[repr(C)]` layout changed and version-3 plugins remain compatible. Reusing `PermissionDenied` for the limit-exceeded case (sanctioned by the issue) avoids adding a `PluginStatus` variant that the exact-match ABI gate would otherwise force a bump for.

## Out of scope

The direct-syscall bypass (an in-process plugin calling `std::net`/`std::fs` directly) — still documented as future work in `core/src/plugin/security.rs`; only OS-level isolation could close it. This is the last planned item in the capability-bridge chain.

## Tests

- Host-side unit tests in `core/src/plugin/capabilities.rs`: ceiling reached → denied and a freed slot re-admits; a failed connect releases its slot; the configured timeout bounds a black-holed dial-out; `ConnectionPolicy::from_manifest` overrides only declared fields.
- Manifest tests: `connectionPolicy` is optional, round-trips, and rejects unknown keys.
- End-to-end test in `core/tests/plugin_capability_bridge.rs` over the real `dlopen` boundary via a new fixture `connlimit` probe: a ceiling of 2 caps 3 attempted connections at 2 allowed / 1 denied.

Closes #2028

### DIFF
--- a/core/src/plugin/capabilities.rs
+++ b/core/src/plugin/capabilities.rs
@@ -21,6 +21,8 @@
 use std::io::Write;
 use std::net::{TcpStream, ToSocketAddrs};
 use std::path::Path;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
 use std::time::Duration;
 
 use termihub_plugin_api::capabilities::{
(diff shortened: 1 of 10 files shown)


armaxri:
Follow-up filed: #2030 — add a dedicated `PluginStatus` for connection-limit refusals (distinct from `PermissionDenied`), which needs an ABI bump and so was deferred out of this PR to keep version 3 stable.

PR #1726, opened 2026-07-19
feat(agent): SSH agent forwarding through the remote-agent SSH backend

armaxri:
Follow-up to #1699 (PR #1718), which delivered SSH agent forwarding (`forwardAgent`) for the desktop russh path only. This brings the same capability to SSH sessions routed through a **deployed termiHub agent**.

## Model decision — agent-host-local agent

The agent already reuses the core SSH backend verbatim (`agent/src/registry.rs` registers `termihub_core::backends::ssh::Ssh`) and runs it in the per-session daemon (`agent/src/daemon/process.rs`). So `forwardAgent` is honored on the agent→target leg by the very same connector request (`auth-agent-req@openssh.com`) and handler bridge that #1699 added — no separate agent-side transport implementation.

Of the two models the issue floated, this PR adopts the **agent-host-local agent** model (and documents it, per the issue's scope): the forwarded-agent channel is bridged to the ssh-agent local to the **agent host** (`$SSH_AUTH_SOCK` / the Windows OpenSSH pipe). The session daemon inherits the agent's environment (`SystemDaemonLauncher` never clears it), so when the **desktop→agent leg is itself SSH with agent forwarding enabled**, the agent host's `$SSH_AUTH_SOCK` already chains to the operator's own agent — the operator's keys reach the final target end to end, transparently, via standard OpenSSH agent chaining rather than a bespoke JSON-RPC relay. No host-local agent is a graceful no-op (reused from #1699).

The heavier JSON-RPC-relay alternative (needed only when the desktop reaches the agent over the TCP `--listen` transport, which has no SSH leg to piggyback) is documented as a limitation and filed as a follow-up.

## Changes

- `agent/src/registry.rs` — doc the chosen forwarding model on the SSH registration (production statement of the model).
- `agent/src/daemon/process.rs` — Rust unit tests pinning the agent-side seam: `forwardAgent` survives the `TERMIHUB_SETTINGS` daemon transport and maps to `SshConfig.forward_agent`; absent → off (graceful default).
- `docs/testing.md` — new section documenting the model + manual verification steps (integration lane only; not per-PR CI).
- Change fragment.

## Test plan

- `cargo test -p termihub-agent daemon_ssh_settings` — new tests pass.
- Connector request, handler bridge and no-agent no-op remain covered by the core tests from #1699.
- Manual end-to-end steps documented in `docs/testing.md` → "SSH agent forwarding through the remote agent".

Closes #1719

Issue #1719, opened 2026-07-19
feat(agent): SSH agent forwarding through the remote-agent SSH backend


armaxri:
Follow-up to #1699 (PR #1718).

#1699 delivered per-connection SSH agent forwarding (`forwardAgent`) for the **desktop russh path** only (`core/src/backends/ssh/**` — the connector requests `auth-agent-req@openssh.com` on the final-target session channel, and `TermiHubHandler::server_channel_open_agent_forward` bridges the server-opened `auth-agent@openssh.com` channel to the local agent). It deliberately left the **remote-agent (termiHub agent) SSH backend** (`agent/src/ssh/`) out of scope.

When a session is reached via a deployed remote agent rather than the desktop russh path, `forwardAgent` currently has no effect on that leg.

## Scope
- [ ] Honor `SshConfig.forward_agent` in the agent's SSH backend (`agent/src/ssh/`), requesting agent forwarding on the session channel it opens.
- [ ] Decide and implement how the agent bridges the forwarded-agent channel: the agent host has no direct access to the *desktop's* local `ssh-agent`, so this likely needs the agent to relay the agent channel back over the desktop↔agent JSON-RPC transport to the desktop's local agent (or an explicit decision that only an agent-host-local agent is used). Document the chosen model.
- [ ] Tests mirroring #1699 (config already carries the field): behavior/validation where feasible, plus manual/integration steps against `tests/docker` SSH containers (integration tests do not run in per-PR CI).

## Done when
- A remote-agent-routed SSH connection with `forwardAgent: true` makes an agent reachable on the target per the documented model, with the no-agent-available case a graceful no-op.

Notes: the `forwardAgent` field, the editor toggle, and the (de)serialization already exist from #1699 — this is purely the agent-side transport. Also related: genuine per-bastion *shell* agent access (termiHub opens no interactive shell on jump hops today) is a separate, larger question noted in `docs/testing.md` and not covered here.

PR #2037, opened 2026-07-26
feat(agent): relay the desktop ssh-agent over the tcp agent transport (Closes #1727)

### LINKED ISSUE #1727: SSH agent forwarding to the desktop's agent over the TCP agent transport
Follow-up to #1719 (PR #1726).

#1719 implemented SSH agent forwarding through the remote-agent SSH backend using the **agent-host-local agent** model: the forwarded-agent channel is bridged to the ssh-agent local to the *agent host* (`$SSH_AUTH_SOCK` / the Windows OpenSSH pipe). When the desktop reaches the agent **over SSH with agent forwarding enabled**, that host-local socket transparently chains back to the operator's own agent, so keys reach the final target end to end — no extra work needed.

## Gap

When the desktop reaches the agent over the **TCP transport** (`--listen`) rather than SSH, there is no SSH leg to piggyback agent forwarding on. In that deployment the operator's local agent is only reachable if the agent host separately runs its own ssh-agent with the right keys.

## Scope

- [ ] Decide whether to relay the desktop's local ssh-agent to the agent host over the desktop↔agent JSON-RPC transport (tunnelling the agent-protocol byte stream), so `forwardAgent` works with the operator's keys even when the desktop→agent leg is TCP.
- [ ] If pursued, wire the relayed stream into the same forwarded-agent bridge the core SSH handler already uses (`core/src/backends/ssh/agent_forward.rs`), so the target sees a normal `$SSH_AUTH_SOCK`.
- [ ] Tests + manual steps against `tests/docker` with the agent reached over `--listen`.

## Done when

An agent reached over the TCP transport with `forwardAgent: true` exposes the **desktop's** agent keys on the final target, with the no-agent case a graceful no-op — matching the SSH-reached behavior from #1719.

Note: this is the larger JSON-RPC-relay model the #1719 issue floated as an alternative; #1719 deliberately shipped the simpler host-local model first.

armaxri:
Follow-up to #1719 (PR #1726), which delivered SSH agent forwarding through a deployed agent using the **agent-host-local** ssh-agent. That model only reaches the operator's keys when the desktop→agent leg is *itself* SSH with forwarding on, so the agent host's `$SSH_AUTH_SOCK` chains back. Over the **TCP (`--listen`) transport** there is no such SSH leg — this PR fills that gap.

## Tunnelling approach (the Concept "decide whether")

Rather than abstracting the core russh bridge, this exploits the fact that the daemon's bridge connects to a filesystem `$SSH_AUTH_SOCK`:

- When an SSH session opts into `forwardAgent` (unix), the agent worker binds a **per-session ssh-agent relay socket** and exports it to the session daemon as `SSH_AUTH_SOCK`. The daemon's existing core SSH bridge (`core/src/backends/ssh/agent_forward.rs`) connects to it exactly as it would a real agent — **"the target sees a normal `$SSH_AUTH_SOCK`"**, unchanged.
- Every connection the daemon makes to that socket is tunnelled byte-for-byte over the desktop↔agent JSON-RPC transport, mirroring the PTY plumbing (`connection.output`/`connection.write`): an id-tagged stream carrying base64 chunks via new `agent.forward.{open,data,close}` messages (protocol **0.5.0**).
- The **desktop** bridges each stream to the operator's **own** local ssh-agent, through the same core connector the russh path uses (`connect_local_agent_boxed`).

No local agent is a graceful no-op (desktop answers `open` with `close`; the daemon's bridge drops the channel) — matching #1719's behaviour. Relaying is **unix-only** for now (a Windows agent host's bridge target is a fixed OpenSSH pipe, not an overridable socket), so Windows keeps the #1719 host-local model; gated by `should_relay`.

## Changes

- `core` — public `connect_local_agent_boxed()` so the desktop reuses the exact russh connect logic.
- `agent` — `session::agent_forward::AgentForwardRelay` (per-session listener + stream registry), wired into daemon spawn (`SSH_AUTH_SOCK` env) and session teardown; new `agent.forward.data`/`agent.forward.close` dispatch methods; protocol bump to 0.5.0.
- `src-tauri` — `terminal::agent_forward::DesktopAgentForward` bridges the tunnelled streams to the operator's local agent; new `AgentIoCommand` variants + I/O-loop routing.
- Docs — `remote-protocol.md` (0.5.0 + method/notification specs), change fragment, `docs/testing.md` manual steps.

## Verification

- Unit tests on all three layers (relay stream open/data/close, dispatch parsing, desktop no-agent no-op).
- **Runtime end-to-end** (`relay_socket_exposes_a_real_ssh_agent_key_end_to_end`): a real `ssh-add -l` connecting to the relay socket sees a **live `ssh-agent`'s** key through a stand-in desktop bridge — proving the target's `$SSH_AUTH_SOCK` exposes the operator's keys over the tunnel. Ran locally against a real ssh-agent; skips where ssh tooling is absent.
- Manual `tests/docker` steps documented (integration lane; not per-PR CI).

Closes #1727

### DIFF
--- a/agent/src/daemon/transport.rs
+++ b/agent/src/daemon/transport.rs
@@ -132,7 +132,8 @@ pub async fn connect(endpoint: &str) -> io::Result<(BoxedReader, BoxedWriter)> {
 #[cfg(unix)]
 #[allow(unused_imports)]
 pub use unix_impl::{
-    endpoint_alive, open_daemon_log, open_registry_log, registry_endpoint, session_endpoint,
+    agent_forward_endpoint, endpoint_alive, ensure_agent_forward_dir, open_daemon_log,
+    open_registry_log, registry_endpoint, session_endpoint,
 };
 #[cfg(windows)]
 #[allow(unused_imports)]
@@ -150,6 +151,19 @@ mod unix_impl {
     pub fn session_endpoint(session_id: &str) -> String {
         socket_dir()
             .join(format!("session-{session_id}.sock"))
+            .to_string_lossy()
+            .into_owned()
+    }
+
+    /// Compute the ssh-agent relay socket path for a session (#1727).
+    ///
+    /// A sibling of the per-session daemon socket in the same per-user `0o700`
+    /// directory. Exported to the session daemon as `SSH_AUTH_SOCK`, it lets the
+    /// daemon's core SSH bridge reach the desktop's agent over the JSON-RPC
+    /// transport as if it were a normal local agent.
+    pub fn agent_forward_endpoint(session_id: &str) -> String {
+        socket_dir()
+            .join(format!("session-{session_id}-agent.sock"))
             .to_string_lossy()
             .into_owned()
     }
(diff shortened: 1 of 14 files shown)


armaxri:
Follow-up filed for the Windows agent-host relay (this PR ships the unix relay only): #2038.

Issue #1849, opened 2026-07-23
fix(terminal): terminal output painted out of order — later prompt/command lines rendered between a previous command's output lines (clears on resize)


armaxri:
Re-scope of the real defect in **image001.png** (originally point 10 of the maintainer's user test). The earlier issue #1828 read that screenshot as a *sidebar* clipping glitch and fixed a real-but-separate sidebar flex bug (PR #1844). The **actual** problem the maintainer reported is in the **terminal output rendering**, and remains unfixed.

## Symptom (from the maintainer, confirmed against image001.png)
Terminal output lines are painted in the wrong order. A `git status` was run, and the **prompt/command lines that came *after* it are displayed *between* the lines of the git status output** instead of after it.

Concretely, in image001.png the `git status` untracked-files block is split:
- `.gitignore`, `README.md` render, then
- **two bare `C:\develop\raspi_info_screen>` prompt lines are inserted mid-list**, then
- the remaining untracked files (`package.json`, `pages/`, `public/`, `scripts/`, `src/`, `systemd/`) continue, then
- `nothing added to commit but untracked files present`.

The two empty prompts should appear *after* the entire `git status` output, not spliced into it.

## Key diagnostic
The maintainer reports **the glitch clears after resizing the tab once** (point 10's original wording). That strongly implies the xterm.js **buffer content is correct** but the **renderer paints rows at stale/wrong positions**, and a resize forces a full re-render that corrects it. So this is a render/refresh-ordering bug, not a PTY/output-ordering bug — though the output write/coalesce path should be ruled out.

Environment: **Windows, local CMD (ConPTY)**, large scrolling output. Reporter is on Windows; the visual may be WebView/renderer-specific.

## Likely area / relationship to other fixes
Probably the same "stale render until resize" family as:
- #1823 (zoomed tab content not visible until scroll — fixed by forcing `xterm.refresh(0, rows-1)` after fit)
- #1828 (sidebar first-paint glitch — fixed separately)

Investigate the terminal write → refresh → scroll path (`src/components/Terminal/`, `TerminalRegistry`, the xterm renderer choice DOM/canvas/webgl, and the write/scroll/refresh batching), and the `core` `OutputCoalescer` (`core/src/output/`, screen-clear detection) to rule out a write-side reorder. Consider whether a deterministic full-refresh after appended output (or on scroll) fixes it, mirroring the #1823 approach — without regressing performance on high-throughput output.

## Scope
[...]

PR #2102, opened 2026-07-27
perf(terminal): add webgl renderer with dom fallback (Closes #2078)

### LINKED ISSUE #2078: Evaluate/add @xterm/addon-webgl renderer (with DOM fallback); scope full-viewport refresh
**Audit — performance (medium).** No GPU-accelerated xterm renderer; DOM renderer plus a forced full-viewport repaint on every flush.
**LOC:** `package.json:44-48` (xterm addons); `src/components/Terminal/Terminal.tsx:711-718,1001-1026`.
**Fix:** Evaluate adding `@xterm/addon-webgl` (with standard context-loss fallback to DOM) — the single biggest render-throughput win. If WebGL is rejected for WebView compat, scope the forced refresh to only rows actually written (where the #1849 paint bug allows). At minimum, document the DOM-renderer choice as an explicit tradeoff.

---
_Filed from the pre-release V&V audit (2026-07-27), "recommended but not blocking" set._

armaxri:
## Summary

Adds GPU-accelerated terminal rendering via `@xterm/addon-webgl` (WebGL2), with the standard context-loss fallback to xterm's DOM renderer. This is the single biggest render-throughput win on high-volume output.

**Outcome of the evaluation: WebGL is viable and shipped.** Verified in a real debug build (`pnpm tauri build --debug`, driven via the system-test bridge) that WebGL2 initializes in the macOS Tauri WKWebView — the terminal container reports `data-terminal-renderer="webgl"`.

## What changed

- **`package.json`** — add `@xterm/addon-webgl@^0.19.0` (pairs with xterm 6).
- **`Terminal.tsx`** — load the WebGL addon after `xterm.open()`. Every failure path degrades to the DOM renderer, never to a blank pane:
  - `loadAddon()` (which runs the addon's `activate()` and creates the WebGL2 context) is wrapped in try/catch — a WebView that cannot provide a context keeps the DOM renderer.
  - `onContextLoss` disposes the addon (xterm reverts to the DOM renderer automatically) on a runtime GPU-context loss.
  - The addon is disposed explicitly on teardown so the GPU context/canvas is released deterministically.
- The active renderer is surfaced on the terminal container as `data-terminal-renderer` (`webgl`/`dom`) — diagnostics for the app-breaking renderer-regression risk, readable in DevTools and by the system-test bridge.
- The forced full-viewport refresh after each flush (#1849 DOM-paint fix) is **left in place** — harmless under WebGL, and required for the DOM fallback path to stay correct.

## Verification (real build, macOS WKWebView)

- Built `pnpm tauri build --debug` and drove it via the Python bridge harness.
- **Renderer active:** the terminal reports `data-terminal-renderer="webgl"` — WebGL2 initialized in the real WebView. No CSP violation blocked it (a CSP-blocked context would throw in `activate()` and report `dom`; it reported `webgl`).
- **Terminal renders:** the shell connects and its content is present/readable (buffer reconstructed via the bridge).
- **Graceful fallback** is covered by unit tests (context loss → DOM, init failure → DOM, no blank pane).

Note: the harness's send-input-then-read-echo path times out on this macOS host, but this reproduces identically with the WebGL addon forced off (DOM renderer) and on the unmodified baseline `test_terminal_runs_commands` — it is a pre-existing macOS bridge-harness limitation (the integration lane runs on Linux/Windows in CI), unrelated to this change.

## Tests

- `Terminal.webgl-renderer.test.tsx` — unit: addon loaded + context-loss handler wired; context loss disposes the addon and flips the diagnostic to `dom`; init failure is swallowed and the DOM path still renders (refresh still issued).
- `tests/system/tests/test_webgl_renderer.py` — integration: the real build reports a valid renderer and renders output (integration lane).
- Full frontend suite green (4665 tests).

Closes #2078

### DIFF
--- a/docs/changes/dev1/perf/2078-webgl-renderer.md
+++ b/docs/changes/dev1/perf/2078-webgl-renderer.md
@@ -1,0 +1,9 @@
+# Changes
+
+## Added
+
+- GPU-accelerated terminal rendering via the `@xterm/addon-webgl` WebGL2
+  renderer, the single biggest render-throughput win on high-volume output
+  (#2078). The renderer degrades gracefully to the DOM renderer if the WebView
+  cannot create a WebGL2 context or the GPU context is lost at runtime, so the
+  terminal never goes blank.
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "@xterm/addon-search": "^0.16.0",
     "@xterm/addon-serialize": "^0.13.0",
     "@xterm/addon-unicode11": "^0.9.0",
+    "@xterm/addon-webgl": "^0.19.0",
     "@xterm/xterm": "^6.0.0",
     "html-to-image": "^1.11.13",
     "lucide-react": "^0.563.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -101,6 +101,9 @@ importers:
       '@xterm/addon-unicode11':
         specifier: ^0.9.0
         version: 0.9.0
+      '@xterm/addon-webgl':
+        specifier: ^0.19.0
+        version: 0.19.0
       '@xterm/xterm':
         specifier: ^6.0.0
         version: 6.0.0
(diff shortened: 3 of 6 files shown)


armaxri:
Follow-up filed: #2107 — skip the per-flush full-viewport refresh when the WebGL renderer is active (the refresh is only needed for the DOM fallback path / #1849).

Issue #2150, opened 2026-07-30
Phase 2: pilot domain — migrate SSH tunnels to Rust store + projector


armaxri:
> ⚠ **BLOCKED on the projection-transport decision — #2139 Open Design Decision #1, "Projection granularity & transport". DO NOT START until the maintainer settles it.** The dispatch/subscribe channel shape this pilot exercises is undecided; do not begin until it is settled and Phase 1 (#2149) has landed.

## Scope

**Phase 2 · Pilot domain.** Migrate one self-contained domain end-to-end onto the new Rust store + projector to prove the pattern. The concept's chosen candidate is **SSH tunnels** — already sliced (its own store slice file), a clear lifecycle, self-contained. Move its authoritative state into a Rust store, project its view model through the Phase 1 channels, convert its mutations to intents, and **delete the tunnel store reducers** on the TS side. Only the tunnel slice changes; everything else is untouched.

## Acceptance criteria

- SSH tunnel authoritative state + lifecycle live in a Rust store; the UI renders the projected tunnel view model and dispatches tunnel intents (create/start/stop/remove).
- The tunnel slice's reducers/authoritative state are removed from `appStore.ts`.
- Projection-assertion bridge tests (from Phase 1) cover the tunnel intents → projections (verify locally; integration tests are not in per-PR CI).
- No behavioural change for tunnels from the user's perspective.

## References

- Concept: `docs/concepts/future/stateless-ui-agent-centric.html` — Phase 2 · Pilot domain (SSH tunnels).
- Part of #2139.

## Dependency / blocker note

**Blocked on Open Design Decision #1 (projection transport) AND on Phase 1 · Substrate (#2149).** Depends on #2149 landing. This pilot is also the prerequisite domain for **S3 · agent-hosted tunnels**.

armaxri:
Unblocked: the projection-transport decision is settled (multi-client in scope → per-region versioned diff channels) and the Phase-1 substrate merged in #2163 (`src-tauri/src/projection/`, `src/services/transport/`, `src/store/projectionCache.ts`). This pilot builds directly on it.

armaxri:
Merged via PR #2167.

PR #2175, opened 2026-07-30
fix(ui): detect desktop shell via __TAURI_INTERNALS__ (Closes #2166)

### LINKED ISSUE #2166: createTransport() isTauri() probe fails in the desktop app (withGlobalTauri off)
Follow-up to #2164 (projection-assertion harness).

## Problem

`createTransport()` in `src/services/transport/index.ts` selects `TauriTransport` only when `isTauri()` returns true, and `isTauri()` checks `"__TAURI__" in window`. The desktop app does **not** set `app.withGlobalTauri` in `src-tauri/tauri.conf.json`, so `window.__TAURI__` is absent even though Tauri IPC is fully available via `window.__TAURI_INTERNALS__` (which `@tauri-apps/api`'s `invoke`/`Channel` use).

Result: inside the running desktop app, `createTransport()` takes the remote-client branch and throws `non-Tauri environment requires a JSON-RPC socket (remote-client mode transport)`.

Discovered while building the #2164 harness: the projection recorder had to construct `TauriTransport` directly instead of via `createTransport()`. This will bite Phase 2 (#2150) the moment a real UI component calls `createTransport()` in the desktop shell.

## Scope
- [ ] Make `isTauri()` (or `createTransport`) detect the desktop shell reliably — e.g. probe `"__TAURI_INTERNALS__" in window` (what `invoke` actually uses), or set `app.withGlobalTauri = true` in `tauri.conf.json`.
- [ ] Add a unit test covering the desktop-detection path.
- [ ] Once fixed, the `ProjectionRecorder` can revert to `createTransport()` (see the note in `src/testbridge/projectionRecorder.ts`).

## Done when
`createTransport()` returns a `TauriTransport` when running in the desktop app, verified by a test, and a real desktop component can use it without the JSON-RPC-socket error.

armaxri:
## Problem

`createTransport()` selected `TauriTransport` only when `isTauri()` returned true, and `isTauri()` probed `"__TAURI__" in window`. The desktop app leaves `app.withGlobalTauri` **off** in `src-tauri/tauri.conf.json`, so `window.__TAURI__` is absent even though Tauri IPC is fully available via `window.__TAURI_INTERNALS__` (what `@tauri-apps/api`'s `invoke`/`Channel` actually use).

Result: inside the running desktop app, `createTransport()` fell through to the remote-client branch and threw `non-Tauri environment requires a JSON-RPC socket`. `ProjectionRecorder` had to construct `TauriTransport` directly to work around it.

## Fix

- `isTauri()` now probes `"__TAURI_INTERNALS__" in window` — the object `invoke`/`Channel` read, so it is the ground-truth gate for whether `TauriTransport` will work. `"__TAURI__"` is kept as a secondary probe for a globals-enabled build.
- Reverted `ProjectionRecorder` to build its transport via `createTransport()` (scope item 3), which now resolves to `TauriTransport` correctly with `withGlobalTauri` off.

## Tests

Added regression tests in `transport.test.ts`:
- `isTauri()` returns true when only `__TAURI_INTERNALS__` is present (desktop, `withGlobalTauri` off).
- `createTransport()` selects `TauriTransport` in that environment.
- A plain browser (no Tauri internals) still falls back to `WebSocketTransport`.

Test commit precedes the fix commit (red → green).

## Verification

Runtime-environment bug — not caught by per-PR CI. Verified in the real desktop app via the projection bridge harness (`test-system-py.sh -m integration -k projection`), which drives the in-app `ProjectionRecorder` → `createTransport()` → `TauriTransport` end to end. Result recorded in a PR comment.

Closes #2166

### DIFF
--- a/src/services/transport/index.ts
+++ b/src/services/transport/index.ts
@@ -31,9 +31,21 @@ export {
 } from "./ProjectionClient";
 export { newClientId, newIntentId } from "./ids";
 
-/** True when running inside the Tauri desktop shell. */
+/**
+ * True when running inside the Tauri desktop shell.
+ *
+ * Probes `__TAURI_INTERNALS__` — the object `@tauri-apps/api`'s `invoke`/
+ * `Channel` actually use — rather than `__TAURI__`. The desktop app leaves
+ * `app.withGlobalTauri` off (see `src-tauri/tauri.conf.json`), so `__TAURI__`
+ * is absent even though IPC is fully available; keying off it would wrongly
+ * fall through to the remote-client path inside the running app (#2166).
+ * `__TAURI__` is kept as a secondary probe for a globals-enabled build.
+ */
 export function isTauri(): boolean {
-  return typeof window !== "undefined" && "__TAURI__" in window;
+  return (
+    typeof window !== "undefined" &&
+    ("__TAURI_INTERNALS__" in window || "__TAURI__" in window)
+  );
 }
 
 /**
--- a/src/services/transport/index.ts
+++ b/src/services/transport/index.ts
@@ -43,8 +43,7 @@ export { newClientId, newIntentId } from "./ids";
  */
 export function isTauri(): boolean {
   return (
-    typeof window !== "undefined" &&
-    ("__TAURI_INTERNALS__" in window || "__TAURI__" in window)
+    typeof window !== "undefined" && ("__TAURI_INTERNALS__" in window || "__TAURI__" in window)
   );
 }
 
--- a/src/services/transport/transport.test.ts
+++ b/src/services/transport/transport.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it, vi } from "vitest";
 
 import { createTransport, isTauri } from "./index";
+import { TauriTransport } from "./TauriTransport";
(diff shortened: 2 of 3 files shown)


armaxri:
## Desktop verification (macOS, real app)

Built the debug desktop app and ran the projection bridge harness against it:

```
./scripts/test-system-py.sh --debug -m integration -k projection
```

Result:

```
tests/test_projection_bridge.py ........
8 passed, 622 deselected in 6.18s
```

The harness drives the running app's projection substrate through the test bridge, whose `ProjectionRecorder` now constructs its transport via `createTransport()` (reverted from the direct-`TauriTransport` workaround). All 8 subscribe → dispatch → diff → gap → resync assertions pass, so `createTransport()` correctly selects `TauriTransport` inside the real desktop shell with `app.withGlobalTauri` off. Before this fix `createTransport()` threw `non-Tauri environment requires a JSON-RPC socket` at the first `subscribe`, which would have failed all 8.

Note: Phase 1 migrates no domain onto the substrate, so the exercised region is the `diag.counter` diagnostic region — there is no live `tunnels` projection region yet (that is Phase 2, #2150). What this bug blocked was the transport selection, which is what is verified here end-to-end.

Issue #2151, opened 2026-07-30
Phase 3: layout & tabs — Rust LayoutStore (activates panelTree port)


armaxri:
> ⚠ **BLOCKED on the projection-transport decision — #2139 Open Design Decision #1, "Projection granularity & transport". DO NOT START until the maintainer settles it,** and until Phase 1 · Substrate (#2149) has landed.

## Scope

**Phase 3 · Layout & tabs — the biggest single win.** Move the panel tree + tab model into a Rust `LayoutStore`. This is where the **Phase 0 `panelTree` port (#2143)** activates: the Rust panel-tree algebra becomes authoritative and the UI renders a flat panel/tab render-list. Drag-drop becomes an intent; visuals stay identical. The `workspaceLayout.ts` helpers used by the restore path also move server-side here.

## Acceptance criteria

- A Rust `LayoutStore` owns the panel tree + tab model, built on the ported `panelTree` algebra (#2143).
- The UI renders a flat panel/tab render-list from the projected layout; drag-drop, split, merge, and tab moves are dispatched as intents.
- Panel-tree authoritative state + reducers removed from `appStore.ts`.
- Visual/behavioural parity with the current layout; projection-assertion tests cover the layout intents.

## References

- Concept: `docs/concepts/future/stateless-ui-agent-centric.html` — Phase 3 · Layout & tabs; Preliminary Implementation Details (`panelTree.ts`/`workspaceLayout.ts` ports activate).
- Part of #2139.

## Dependency / blocker note

**Blocked on Open Design Decision #1 (projection transport).** Depends on **Phase 1 · Substrate (#2149)** and on the **Phase 0 `panelTree` port (#2143)** landing (activates it). Do not start until decision #1 is settled and #2149 + #2143 are merged.

armaxri:
**Unblocked.** The projection-transport decision is settled (multi-client in scope → per-region versioned diff channels), and both prerequisites merged: Phase-1 substrate (#2163) and the Phase-0 `panelTree` port (#2143). The ⚠ BLOCKED note above is now stale — this is ready to implement. The projection-assertion harness (#2164) is available for the layout-intent tests.

armaxri:
## Phase 3 scoping — blocked on Decision #2, and needs decomposition (analysis before implementing)

[...]

PR #2177, opened 2026-07-30
feat(ui): cut structural layout mutations to the store (Part of #2151)

armaxri:
## Phase 3 step 2 — cut structural layout mutations over to the store

`Part of #2151` (the umbrella stays open for steps 3–5).

Step 1 (PR #2174) landed the shadow, backend-authoritative `LayoutStore` +
client-scoped `layout@<clientId>` region + `layout.*` intents, but nothing in
the UI touched it. Step 2 starts making the store authoritative for
**panel-tree structure**: structural mutations dispatch `layout.*` intents and
the resulting projection diff is reconciled back into `appStore`'s existing
panel-tree shape, so `SplitView` renders exactly as before (rendering is cut in
step 3, the reducers removed in step 4).

### What this does

- **Backend — `layout.replace` seed intent.** The shadow store is never fed tabs
  (tab *creation* stays in `appStore` under partial projection), so the four
  structural intents can't operate on the real tree. `layout.replace` installs a
  client's whole tree (minimal-tab form). Also aligns the store's
  `with_tab_removed` to the frontend's **positional** active-tab fallback
  (`min(idx, len-1)`), so a source panel focuses the same tab after a move.
- **Frontend — the projection bridge (`src/store/layoutBridge.ts`).** For each cut
  mutation it *seeds* the store with `appStore`'s current tree, dispatches the
  intent, awaits the diff, and **reconciles** the minimal-tab view back into rich
  `TerminalTab`s by id. Because the xterm DOM is registered globally by tab id
  (`TerminalRegistry`) and `TerminalSlot` reparents it, a moved tab keeps its
  **live session + scrollback**. Seeding per-op means the store never drifts from
  `appStore` between ops.
- **Cut so far:** `splitPanel` → `layout.split`; `splitPanelWithTab` (drag to
  panel center = merge a tab into the stack, drag to edge = split) →
  `layout.moveTab`. The single-tab **self-edge-drop** corner (where the store
  collapses the emptied source but the local reducer keeps it) stays on the local
  path to preserve exact parity.

### Strangler safety

Gated by `layoutIntentsEnabled()` — **off by default**, so the shipped app keeps
running on the untouched local reducers; even when enabled, any
dispatch/reconcile failure falls back to the local mutation. Rendering and the
`appStore` reducers are unchanged (steps 3–4). The flag is runtime-flippable
(`localStorage["termihub.layoutIntents"]` / `window.__TERMIHUB_LAYOUT_INTENTS__`)
so step 3 can flip the default after a GUI verification pass.

### Left for follow-ups (filed, `Part of #2151`)

- `moveTab` (tab-bar drop at an index) needs an **indexed** `layout.moveTab`
  payload the backend intent doesn't yet carry.
- `closeTab` structural half — session/side-map coupling warrants a focused,
  atomic cut.
- `reorderTabs` intra-panel reorder.

### Tests

- Rust: `layout.replace` store + projection round-trip; positional
  active-tab-removal parity. (25 layout tests green.)
- Frontend: bridge unit tests (rich⇄minimal map, reconcile re-hydration by id,
  unknown-tab throw, flag) + `appStore` round-trip (seed→intent→diff→appStore,
  flag-off, rejection-fallback) + **cut↔local parity comparisons** running
  `splitPanel`, drag-center, drag-edge, and a middle-active-tab move both ways and
  asserting identical trees.

### Verification note

The interactive GUI drag-drop pass (mouse drags + visual scrollback check) was
**not** performed here; it is substituted by the automated cut↔local parity tests
above, and the flag ships **off** so there is zero user-facing change. Recommend a
manual `./scripts/dev.sh` pass with the flag on before step 3 flips the default.

### DIFF
--- a/src-tauri/src/layout/projection.rs
+++ b/src-tauri/src/layout/projection.rs
@@ -27,6 +27,7 @@
 //! | `layout.merge`              | `{ sourcePanelId, targetPanelId }`       | move all tabs into the target, drop the source |
 //! | `layout.moveTab`            | `{ tabId, targetPanelId, edge }`         | move a tab (center = merge; edge = split)      |
 //! | `layout.closeTabStructure`  | `{ tabId }`                              | remove a tab; drop its leaf if left empty      |
+//! | `layout.replace`            | `{ root, activePanelId }`                | install a whole tree (the bridge's seed path)  |
 //!
 //! # Shadow mode
 //!
@@ -42,6 +43,8 @@ use std::sync::Arc;
 use serde::de::DeserializeOwned;
 use serde_json::Value;
 use tauri::{AppHandle, Manager};
+
+use termihub_core::layout::panel_tree::PanelNode;
 
 use crate::layout::store::{LayoutError, LayoutStore};
 use crate::projection::{HandlerRegistry, Intent, ProducedRegion, Projector};
@@ -109,7 +112,7 @@ pub fn register_layout_intents(registry: &mut HandlerRegistry, app_handle: AppHa
         Ok(publish_layout(projector, &store, &intent.client_id))
     });
 
-    let handle = app_handle;
+    let handle = app_handle.clone();
     registry.route("layout.closeTabStructure", move |intent, projector| {
         let store = store_of(&handle)?;
         let tab_id = required_str(intent, "tabId")?;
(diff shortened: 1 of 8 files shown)


armaxri:
Follow-ups filed (all `Ready2Implement`, Part of #2151):
- #2178 — cut indexed `moveTab` (tab-bar drop) to the store (needs an indexed `layout.moveTab` payload)
- #2179 — cut `closeTab`'s structural half (atomic with its session/side-map teardown)
- #2180 — cut `reorderTabs` (needs a new `layout.reorderTabs` intent)

Issue #2206, opened 2026-07-30
Phase 4 step 5: move restore-cohort + broadcast/workflow state machines server-side


armaxri:
Follow-up to #2152 (PR #2202). Phase 4 step 5 of #2139 — the restore/broadcast/workflow half of #2152's acceptance that the bounded shadow (PR #2202) deliberately deferred to keep a clean per-domain boundary. **Blocked on step 4**.

## Scope
- [ ] Move the startup **restore-cohort** state machine server-side, built on the ported #2145 `restore_mode` engine (`termihub_core::restore_mode` becomes authoritative), including the bulk "Reconnect failed tabs" cohort (#1227) as projected state + intents.
- [ ] Move the **broadcast** membership + **workflow** state machines server-side onto the projection substrate.
- [ ] Decide region scope for the cohort/broadcast state (shared vs client-scoped) against Decisions #4/#6 and record it, as PR #2202 did for `session-lifecycle`.

## Done when
- Restore/broadcast/workflow state machines live server-side on the ported #2144/#2145 logic; the UI renders projected status and issues intents.
- Projection-assertion tests cover the cohort transitions; no user-visible change to restore/broadcast behaviour.
- #2152's acceptance criteria are fully met → #2152 can close.

Blocked on step 4.

armaxri:
Unblocked — the session cut is live by default (#2233) and step 4 (#2205) is reframed to keep-reducers-as-fallback, so this (restore-cohort + broadcast/workflow state machines server-side) can proceed. Sequence after or alongside #2205.

armaxri:
PR #2240 lands the **restore-cohort** shadow foundation (client-scoped `restore-cohort@<clientId>` region + `restore.*` intents on the substrate; zero user-facing change). Region-scope justified in the module docs per Decision #4/#6, as #2202 did for session-lifecycle.

Remaining scope split into `Ready2Implement` follow-ups so each stays a reviewable per-domain slice:
- #2241 — restore-cohort cut-over (render-cut → flag-gated mutation-cut, reducers kept as fallback)
- #2242 — broadcast membership machine
- #2243 — workflow-run machine

Keeping this issue open until those land; #2152 closes when all three are server-side.

armaxri:
[...]

Issue #2164, opened 2026-07-30
Projection-assertion bridge harness for the stateless-UI substrate


armaxri:
Follow-up to #2149 (PR #2163). Phase 1 delivered the projection substrate (dispatch/subscribe channels + per-region versioned diff fan-out) with Rust and TS unit harnesses. The **Python bridge** projection-assertion harness — the deferred half of Phase 0's test scaffolding — was left out of Phase 1 because it needs a region actually wired into the running app to assert against, and the integration lane it runs in does not execute in per-PR CI.

## Scope
- [ ] In `tests/system/`, drive an `intent_dispatch` over the WebSocket bridge and subscribe to the affected region.
- [ ] Assert the pushed projection frame(s): snapshot-on-attach, then the diff produced by the intent (baseVersion/version + ops).
- [ ] Cover a forced gap -> resync re-baseline over the bridge.
- [ ] Requires a concrete region wired into the app to exercise (either the first migrated domain from Phase 2 #2150, or a minimal diagnostic region). Coordinate with whichever lands first.
- [ ] Verify with a **local** container run (`./scripts/test-system-py.sh -m integration`) — integration tests do not run in per-PR CI.

## Done when
The Python bridge asserts a backend-pushed view model for a dispatched intent, and the gap->resync path, against a region wired into the running app.

Issue #2152, opened 2026-07-30
Phase 4: session lifecycle state machines move server-side


armaxri:
> ⚠ **BLOCKED on the projection-transport decision — #2139 Open Design Decision #1, "Projection granularity & transport". DO NOT START until the maintainer settles it,** and until Phase 1 · Substrate (#2149) has landed.

## Scope

**Phase 4 · Session lifecycle.** Move the connect / reconnect / restore / broadcast / workflow state machines server-side; the UI renders projected session status. The real sessions already live backend-side — only the state machine relocates. This is where the **Phase 0 `reconnectBackoff` (#2144)** and **`restoreMode` (#2145)** Rust ports become authoritative (backoff + restore-cohort logic).

## Acceptance criteria

- Connect/reconnect/restore/broadcast/workflow state machines live server-side, built on the ported `reconnectBackoff` (#2144) and `restoreMode` (#2145) logic.
- The UI renders projected session status via subscriptions and issues lifecycle intents; the corresponding authoritative state + reducers are removed from `appStore.ts`.
- Projection-assertion tests cover the lifecycle transitions.
- No user-visible change to connect/reconnect/restore behaviour.

## References

- Concept: `docs/concepts/future/stateless-ui-agent-centric.html` — Phase 4 · Session lifecycle; Phase 0 · Test scaffolding (backoff, restore cohort state machines).
- Part of #2139.

## Dependency / blocker note

**Blocked on Open Design Decision #1 (projection transport).** Depends on **Phase 1 · Substrate (#2149)**; consumes the Phase 0 ports **#2144** (reconnectBackoff) and **#2145** (restoreMode). Do not start until decision #1 is settled and #2149 is merged.

armaxri:
## Phase 4 step 1 landed as a bounded shadow (PR #2202)

Following the Phase-3 template, step 1 is a **shadow, backend-authoritative session-lifecycle state machine** — zero user-facing change, nothing subscribes yet:

- New `src-tauri/src/session_projection/`: a `SessionLifecycleStore` modelling the connect/reconnect/disconnect/error machine the frontend drives per session, composing the **ported #2144 `reconnect_backoff`** engine for the retry sub-loop.
- A single **shared** `session-lifecycle` region (Open Design Decision #4 — infrastructure domains are shared; a session's status is a property of the session, not the viewing client, exactly like tunnels). Per-client presentation (dismissed overlay, view-mode) stays frontend, keeping this a clean per-domain region.
[...]

Issue #2139, opened 2026-07-29
Stateless UI / agent-centric redesign: move all logic to backend + run services/tools on the agent


armaxri:
## Summary

A long-horizon architecture concept to restructure termiHub around two convergent principles:

1. **Stateless UI** — move *all* business logic and authoritative state out of the React
   frontend into the Rust backends (`src-tauri` desktop + `termihub-core` + the remote `agent`).
   The UI becomes a thin, stateless projection/dispatcher: it renders backend-provided view state
   and dispatches intents, holding no authoritative state or business logic of its own (only
   view-local ephemera like scroll/focus/hover).

2. **Agent-centric services & tools** — allow *all* services and network tools to be executed by
   the remote `agent`, not only the desktop: the embedded servers (HTTP/FTP/TFTP), the network
   utilities (ping, traceroute, port scanner, DNS, HTTP monitor, Wake-on-LAN), tunnels, and
   monitoring — behind one **service/tool trait** that both the desktop and the agent host.

## Why now

Survey of the current tree (see the concept): ~45-55% of the 80.8k lines of non-test production
frontend TypeScript is business-logic/state, concentrated in a single **8,353-line `appStore.ts`**
plus large logic-bearing components. Meanwhile the shared `core/` crate already hosts the
`ConnectionType`, `MonitoringProvider`, and `FileBrowser` traits, the network diagnostics, and the
IPC substrate — and the **agent already exposes `network.*` diagnostics** and proxies files +
monitoring. The foundations for both principles already exist; this concept designs the full
target and a phased, strangler-style migration that keeps the app working throughout.

## Deliverable

Design concept (speculative / long-horizon → `future/` tier):
`docs/concepts/future/stateless-ui-agent-centric.html` — current-architecture map, target
stateless-UI + agent-centric design (state-flow / intent-flow / service-tool-execution Mermaid
diagrams), the agent-executed-services/tools design, a phased migration path, risks, and open
decisions.

This is a **design-first** tracking issue — no implementation is in scope here. A full
redesign/restructure is explicitly accepted by the maintainer as the north star; individual
migration phases will be filed as their own implementable issues.

armaxri:
[...]

PR #2240, opened 2026-07-31
feat(backend): shadow restore-cohort projection authority (Part of #2206)

armaxri:
## What

Phase 4 **step 5** of the stateless-UI migration (#2206, part of #2152 / #2139) — the first, highest-value coherent slice: a **shadow, backend-authoritative restore-cohort state machine**, mirroring how step 1 (#2202) landed the session-lifecycle shadow.

Moves the startup restore/launch **cohort-aggregation** machine the frontend drives (`appStore` `restoreCohort` + `failedRestoreTabIds`, #1146 / #1227) into a Rust authority on the projection substrate:

- New `src-tauri/src/restore_cohort_projection/`: a `RestoreCohortStore` that, per client, registers the tabs one restore placed (`beginCohort`), settles them as each connects or fails (`settleTab`), and on the last settle raises one summary plus the failed-tab set that drives the bulk **"Reconnect failed tabs"** control (#1227).
- Served via `restore.*` intents (`restore.beginCohort`, `restore.settleTab`) on the substrate dispatcher, with state-machine + projection-assertion tests (#2164 harness) — 20 tests.

### Region scope (Decision #4 / #6 — recorded as #2202 did)

**Client-scoped** `restore-cohort@<clientId>`. Unlike the shared `session-lifecycle` region (a session's status is a property of the shared session), a restore cohort is an **orchestration overlay + aggregate feedback owned by the client that launched the restore** — each window restores its own layout, raises its own summary toast, and retries its own failed tabs. It is a property of the viewing client, so it mirrors the client-scoped `layout` region.

### Shadow mode — zero user-facing change

Registered and fully served, but nothing in the live UI subscribes to or renders a `restore-cohort` region and no frontend code dispatches `restore.*` yet. The `appStore` cohort reducers and the toast feedback remain authoritative. Two concerns deliberately stay frontend for the later render-cut, keeping a clean per-domain boundary: the **toast render** itself (the store projects a `settlement` summary carrying a monotonic `seq`), and the **live-terminal filter** (restricting the retry set to still-live terminal tabs needs the tab registry that lives frontend-side).

## Scope of #2206 remaining (follow-ups filed)

#2206 stays **open** until all three machines are server-side. This PR lands restore-cohort's shadow; the rest is filed as `Ready2Implement` follow-ups referencing #2206:

- restore-cohort **cut-over** (render-cut, then flag-gated mutation-cut keeping the reducers as the parity-safe fallback)
- **broadcast** membership machine (shadow → cut)
- **workflow** run machine (shadow → cut)

Part of #2206
Part of #2152

## Tests

- `cargo test -p termihub --lib restore_cohort_projection` — 20 passing (11 state-machine + 6 projection-contract, plus the substrate parity/isolation cases).
- `cargo clippy -p termihub --lib --all-features -- -D warnings` clean; `cargo fmt --check` clean.
- No frontend changes, so the frontend suite is unaffected.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -41,6 +41,13 @@ mod network;
 /// versioned diff channels with multi-subscriber fan-out. Public so integration
 /// tests can drive the projector directly.
 pub mod projection;
+/// Shadow restore-cohort authority (#2206, Phase 4 step 5 of #2139): the
+/// client-scoped `restore-cohort@<clientId>` projection region + `restore.*`
+/// intents, modelling the startup restore/launch cohort aggregation (#1146 /
+/// #1227) on top of the ported restore-decision engine (`termihub_core::restore_mode`,
+/// #2145). Registered and served but not yet driving the live UI — see
+/// [`restore_cohort_projection`].
+mod restore_cohort_projection;
 pub mod run_location;
 mod session;
 mod session_history;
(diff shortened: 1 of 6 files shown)


armaxri:
Follow-ups filed for the remaining #2206 scope (all `Ready2Implement`):
- #2241 — restore-cohort cut-over (render + flag-gated mutation-cut, keep reducers as fallback)
- #2242 — broadcast membership machine server-side
- #2243 — workflow-run machine server-side

#2206 stays open until 5a/5b/5c land; #2152 closes with them.

Issue #2136, opened 2026-07-29
Sandbox frontend plugin execution and drop blob: from script-src (follow-up to #2059)


armaxri:
Follow-up to #2059 (which tightened the CSP and added a CSP-violation CI guard, but deferred the plugin sandbox). This is the residual security win from the original #2059 scope.

## Context
Frontend (JavaScript) plugins are still injected into the **main WebView** document: the loader (`src/plugins/frontendPlugins.ts`) reads the plugin's JS and appends it as `<script src="blob:…">`. That runs plugin code with full main-context/IPC access, and is the sole reason the production CSP still carries `blob:` in `script-src`:

```
script-src 'self' blob: 'wasm-unsafe-eval'
```

Frontend plugins remain **experimental and default-off** (`frontendPluginsEnabled`, #2048), so this is gated — but the residual `blob:` allowance and the weak-isolation model should be closed.

## Why it was deferred from #2059
A correct sandbox is substantial and touches the plugin API surface:

- **Protocol parsers** (`transform(data, sessionId)`) run **synchronously** in the terminal output hot path (`src/components/Terminal/Terminal.tsx` → `transformOutput`). Moving execution out of the main context (Worker/iframe) makes `transform` **async**, which requires an ordered per-session async pipeline in the terminal renderer — the most ordering/perf-sensitive code in the app.
- **Status-bar widgets** (`render(): HTMLElement` / `dispose()`) mount their DOM directly into the main status bar (`src/components/StatusBar/PluginStatusBarWidgets.tsx`). An opaque-origin sandbox (Worker has no DOM; a `sandbox="allow-scripts"` iframe has an opaque origin) cannot hand a live main-document element back — the widget DOM must be **re-homed** across the boundary (per-widget embedded iframe, or a declarative widget model), which is an API change.
- The `plugin-system` concept (`docs/concepts/implemented/plugin-system.html`) currently documents weak isolation as accepted and calls Workers/iframes an explicit **future improvement** — so the concept must be updated as part of this work (it is the source of truth).

## Scope
- [ ] Execute frontend plugin JS in a real sandbox with an opaque origin (a same-origin bootstrap asset loaded into a `sandbox="allow-scripts"` iframe, or a Worker where DOM is not needed), with an explicit `postMessage` API — no plugin JS in the main document, no IPC access.
- [ ] Make the protocol-parser pipeline async with preserved per-session ordering; keep the no-plugin path synchronous and untouched (the fast path).
[...]

Issue #2059, opened 2026-07-27
Sandbox frontend plugins and remove residual CSP allowances (follow-up to #2048)


armaxri:
Follow-up to #2048 (PR #2058). #2048 shipped a restrictive CSP and gated frontend (JavaScript) plugins behind an experimental, default-off opt-in. Two residual CSP allowances remain that a stronger isolation model would let us drop.

## Context
- Frontend plugins are currently injected into the **main WebView** (weak isolation) and run with full IPC access. To load under the strict CSP without `'unsafe-inline'`, plugin code is loaded from a **blob URL**, so `script-src` includes `blob:`.
- The system-test bridge drives a real `tauri build` artifact (CSP enforced) and its in-app client opens `ws://127.0.0.1:<port>` out to the runner, so `connect-src` allows **loopback** WebSockets.

## Scope
- [ ] Execute frontend plugins in a real sandbox (Web Worker or sandboxed iframe) with an explicit message-passing API instead of injecting into the main WebView, enabling per-plugin permission enforcement.
- [ ] Once plugins no longer need runtime script injection into the main document, **remove `blob:` from `script-src`** (and re-evaluate `worker-src`/`child-src`).
- [ ] Move the test bridge off an in-WebView WebSocket (e.g. a Tauri IPC channel / custom protocol) **or** apply a test-only CSP override at build time, then **remove the `ws://127.0.0.1:* ws://localhost:*` allowance** from the production `connect-src`.
- [ ] Add an automated (non-macOS) CI check that asserts the production WebView boots and a terminal renders under the shipped CSP, so future CSP regressions are caught.

## Done when
- Frontend plugin JS no longer executes in the main WebView document context.
- The production CSP contains neither `blob:` in `script-src` nor any `ws://` allowance.
- A CI check verifies the app boots + terminal renders under the CSP.

armaxri:
PR #2137 ships a **safe subset** of this issue and does not close it:

- ✅ Removed the `ws://127.0.0.1:* ws://localhost:*` allowance from the **production** `connect-src` (re-added for test builds only via a `--config` overlay).
- ✅ Added a CSP-violation reporter + an integration guard (`test_csp.py`) that boots the built app, renders a terminal, and asserts zero CSP violations under the shipped CSP, plus a per-PR static CSP-invariant test.

[...]

Issue #2048, opened 2026-07-27
Ship a restrictive CSP and gate frontend plugins as experimental for v0.1.0


armaxri:
**Audit finding — security (medium, cluster).** `tauri.conf.json` sets `security.csp: null` while third-party plugin `frontend/index.js` is injected as inline script into the **main** WebView, and frontend plugins run with full IPC/command access and no per-plugin permission enforcement. Together this turns the accepted weak-isolation model into a real exploit surface.

**Locations:** `src-tauri/tauri.conf.json:24-27` (csp: null); injection `src/plugins/frontendPlugins.ts:46-51,104-110`; posture `src/plugins/pluginRuntime.ts:26-28`.

**Fix:**
- Define a restrictive CSP (e.g. `default-src 'self'; connect-src 'self' ipc: http://ipc.localhost; img-src 'self' data:; object-src 'none'; frame-src 'none'`), accommodating inline plugin scripts via a nonce or the Tauri isolation pattern.
- For v0.1.0, gate frontend-plugin execution behind an explicit **experimental/opt-in** setting with prominent trust UX, and track per-plugin permission enforcement (ties to #2001).

---
_Filed from the pre-release V&V audit (multi-agent, 2026-07-27). Medium/low findings were flagged by the audit but not independently adversarially re-verified — confirm while fixing._

PR #2250, opened 2026-07-31
feat(ui): sandbox frontend plugin execution in a web worker

armaxri:
Part of #2136.

Moves frontend (JavaScript) plugin **execution** out of the main WebView document into a least-privilege **Web Worker sandbox**, closing the weak-isolation gap left by the restrictive-CSP work (#2059). Builds on the PR-0 baseline (#2248), which stays green.

## What changed

Frontend plugin JS no longer runs in the main document (it was injected as a `<script src="blob:…">`). It now runs inside a classic Web Worker with **no DOM, no `window`, and no Tauri IPC** — so plugin code cannot reach the host app, the document, or the backend. Host and worker communicate only over `postMessage`.

- **Protocol parsers** run in the worker. The terminal output path is now an **ordered asynchronous pipeline**: each chunk takes a per-session FIFO slot, the transform runs in the worker, and the host drains from the head so results always reach xterm in arrival order regardless of resolution order. The **no-plugin fast path stays fully synchronous and byte-exact** — when no parser is registered (the default), the terminal never does a sandbox round-trip, so the common case is untouched. Liveness guards (per-session backlog cap + head-of-line watchdog) mean a runaway plugin can only occupy the worker, never freeze the UI thread (strictly better than before).
- **Status-bar widgets**: a Worker has no DOM, so `render()` now returns a serialisable descriptor (`{tag, text, attrs, children}`) instead of a live element. The host materialises it through a strict tag/attribute allowlist — **never `innerHTML`** — so plugin content can't inject executable markup. `dispose()` runs in the worker.

### Why Worker, not iframe
`worker-src 'self' blob:` is already in the CSP, so **this slice makes zero CSP changes** — `frame-src`/`object-src` stay `'none'` (both locked by `cspConfig.test.ts`; an iframe would have required unlocking `frame-src`, exactly the CSP-regression class of #2048/#2059/#2083). Plugin code loads into the worker from a `blob:` URL via `importScripts`, confined to the worker. Monaco's own workers already load under this exact CSP in production, so the substrate is proven.

### `blob:` stays (deferred)
`script-src` keeps `blob:` — plugin code still enters the worker via a blob URL. Dropping `blob:` from `script-src` is the **final** step of #2136 and is filed as a follow-up (see below). The CSP invariant test adds a lock on the worker-src/child-src substrate but does **not** assert the blob drop.

## Files
New sandbox modules under `src/plugins/sandbox/` (protocol, worker runtime core, worker entry, host, widget materialiser, widget store) + unit tests. Rewired: `frontendPlugins.ts` (loader → sandbox host), `Terminal.tsx` (async pipeline), `PluginStatusBarWidgets.tsx` (descriptor rendering). Removed the old main-context `pluginRuntime.ts`. Concept §13 updated to the new isolation model. Baseline harness fixture widget `render()` moved to the descriptor form (no `document` in a worker).

## Verification
- `pnpm build` (tsc + vite) green; the worker bundles as a classic `'self'`-origin chunk using `importScripts` — no CSP change needed.
- Full frontend suite green (4823 tests), incl. new sandbox coverage: ordered drain under out-of-order results, byte-exact pass-through, worker lifecycle, backpressure/watchdog, widget relay, allowlist/no-innerHTML.
- Lint + prettier clean.
- The PR-0 baseline integration test (`test_plugin_sandbox.py`) — enabled plugin's widget renders + parser transforms + **zero CSP violations** under the enforced production CSP — was run locally against a built app. _(result noted in a PR comment.)_
- No Rust changes.

## Follow-up
- Drop `blob:` from `script-src` entirely (the deferred last step of #2136) — filed as #2251.


armaxri:
**Local integration verification** (the `test_plugin_sandbox.py` baseline runs only on the integration lane, not per-PR CI, so verifying locally against a production build was required):

Built the app with the enforced production CSP (`./scripts/test-system-py.sh --debug -m integration -k plugin_sandbox`) and ran the PR-0 baseline — **4/4 passing, reproducibly**:
- seeded plugin reaches `active` but stays dormant while the gate is off;
- enabling the gate renders the plugin's status-bar widget with **zero CSP violations** under the enforced CSP;
- the plugin's protocol parser transforms live terminal output (`sbxparsersrc`→`sbxparserhit`) — proving the sandbox Worker round-trip works end-to-end;
- disabling the gate live unmounts the widget.

The first build surfaced a real bug this integration test caught (and per-PR CI could not): with a parser active, transformed output is pushed to the render buffer from a Worker `message` turn, and a `requestAnimationFrame` armed there is not reliably serviced by the WebView (no paint scheduled) — so output never flushed to xterm. Fixed in `fix(terminal): flush plugin-transformed output when raf is starved` by pairing the RAF with a `setTimeout` fallback; the no-plugin fast path is unaffected. The baseline is green after the fix.

Issue #2227, opened 2026-07-30
Phase 5: migrate the Settings domain to a Rust store + projection region


armaxri:
Part of #2153 (Phase 5 · the long tail) and #2139 (stateless-UI port).

Migrate the **Settings** domain (`AppSettings` — the persisted app-settings slice of `appStore.ts`) onto a Rust store + projection region, following the proven per-domain pattern (Phases 3 & 4).

### Region scope (Decision #4)
**Shared** — `AppSettings` is a single persisted document. In a multi-client world every client sees the same settings and an edit projects to all. (Any device-local preference that should NOT sync would be carved out as client-scoped; call it out at shadow time.)

### Scope
- [ ] **Shadow**: `src-tauri/src/settings_projection/` — a store wrapping the authoritative `AppSettings` (already persisted in Rust) + a `settings` region + `settings.*` intents (update/patch, reset), UI NOT cut over. Projection-assertion tests (#2164 harness). Zero user-facing change.
- [ ] **Render cut**: the Settings UI + settings-driven behavior read the projected `settings` region (parity-safe).
- [ ] **Mutation cut**: flag-gated — `saveSettings`/patch dispatch `settings.*` intents.
- [ ] **Remove reducers**: delete the `AppSettings` appStore state + reducers once the region is authoritative.

### Done when
`appStore.ts` no longer holds settings authoritative state; the Settings UI renders from the `settings` projection; projection-assertion tests cover the domain; nothing user-facing changed at each step.

### Sequencing
Independent. Settings changes fan out widely across the UI, so the render cut is the delicate step — sequence its parity verification carefully. Many other domains read settings, so land this after the more self-contained domains to avoid churn.

armaxri:
**Render cut landed** — PR #2270 (merged into \`develop\`).

The Settings **screen** now renders the persisted \`AppSettings\` document from the shared \`settings\` projection region, via the new \`settingsBridge\` + \`useProjectedSettings()\` hook (faithful-mirror gate, appStore fallback, render flag on by default). Parity-safe, no user-visible change.

Because \`AppSettings\` has ~40 render readers, this cut a coherent bounded slice — the Settings screen panels (\`src/components/Settings/*\`). The remaining **settings-driven behavior** readers (Terminal, StatusBar, NetworkTools, FileEditor, ConnectionEditor, etc.) are tracked in the follow-up **#2269** (Ready2Implement).

[...]

Issue #2226, opened 2026-07-30
Phase 5: migrate the Agents domain to a Rust store + projection region


armaxri:
Part of #2153 (Phase 5 · the long tail) and #2139 (stateless-UI port).

Migrate the **Agents** domain (remote-agent definitions, folders, live agent connections and agent-hosted sessions — the agent slice of `appStore.ts`) onto a Rust store + projection region, following the proven per-domain pattern (Phases 3 & 4).

### Region scope (Decision #4)
**Shared** for agent connection/session status (infrastructure, like tunnels/session-lifecycle: a property of the agent, not the viewer). Agent **definitions/folders** are shared persisted config. Confirm at shadow time.

### Scope
- [ ] **Shadow**: `src-tauri/src/agents_projection/` — a store modeling agent definitions/folders + live connection/session status + an `agents` region + `agent.*` intents (connect/disconnect/shutdown/applySettings/save/update/delete definition, folder CRUD), UI NOT cut over. Projection-assertion tests (#2164 harness). Zero user-facing change.
- [ ] **Render cut**: the agent sidebar + Open Connections read the projected `agents` region (parity-safe).
- [ ] **Mutation cut**: flag-gated — agent mutations dispatch `agent.*` intents.
- [ ] **Remove reducers**: delete the agent appStore state + reducers once the region is authoritative.

### Done when
`appStore.ts` no longer holds agent authoritative state; the agent sidebar and Open Connections render from the `agents` projection; projection-assertion tests cover the domain; nothing user-facing changed at each step.

### Sequencing
Independent of Monitors/Connections/Settings. Larger surface (definitions + folders + live status) — consider splitting the shadow across two PRs (definitions/folders vs live status) if it grows past reviewable size.

armaxri:
Render cut landed in PR #2252 (merged into `develop`).

The agent sidebar (`ConnectionList`, `AgentNode`) and Open Connections now render the agent list, connection status and per-agent sessions/definitions/folders from the shared `agents` projection region via `useProjectedAgents`, falling back to `appStore` whenever the region is not a faithful mirror — parity-safe, no user-visible change. Added an `agent.replace` seed intent (analog of `monitor.replace`) plus `agentsBridge` (flag `agentRenderFromProjectionEnabled`, on by default, runtime-overridable). `appStore` agents fields remain in place.

[...]

Issue #2224, opened 2026-07-30
Phase 5: migrate the Monitors domain to a Rust store + projection region


armaxri:
Part of #2153 (Phase 5 · the long tail) and #2139 (stateless-UI port).

Migrate the **Monitors** domain (SSH/session system-monitoring stats — the `monitors: Record<MonitorKey, MonitoringEntry>` + `monitoringStatsCache` slice in `appStore.ts`) onto a Rust store + projection region, following the proven per-domain pattern (Phases 3 & 4).

### Region scope (Decision #4)
**Shared** region `system-monitors`. A monitor's stats/status/paused/interval are properties of the backend session's `MonitoringProvider`, not of a viewing client — two clients viewing the same session see the same stats. Consistent with the tunnels and session-lifecycle infrastructure regions.

### Scope
- [ ] **Shadow**: `src-tauri/src/system_monitor_projection/` — a `SystemMonitorStore` modeling the authoritative monitor map (reusing `termihub_core::monitoring::{SystemStats, MonitorStatus}`) + a shared `system-monitors` region + `monitor.*` intents defined and registered, UI NOT cut over. Projection-assertion tests (#2164 harness). Zero user-facing change. *(this issue's PR)*
- [ ] **Render cut**: the status bar + Open Connections read the projected `system-monitors` region (parity-safe).
- [ ] **Mutation cut**: flag-gated — `connectMonitoring`/`disconnectMonitoring`/`setMonitoringPaused`/`setMonitoringInterval`/`cancelMonitoring` dispatch `monitor.*` intents instead of driving the appStore slice.
- [ ] **Remove reducers**: delete the `monitors`/`monitoringStatsCache` appStore state + reducers once the region is authoritative.

### Done when
`appStore.ts` no longer holds monitor authoritative state; the status bar and Open Connections render from the `system-monitors` projection; projection-assertion tests cover the domain; nothing user-facing changed at each step.

### Sequencing
Independent of the other long-tail domains (no shared files). Good template-setter — take first.

armaxri:
Shadow foundation landed via #2230 (merged): the `SystemMonitorStore` + shared `system-monitors` projection region + `monitor.*` intents, zero user-facing change (backend-only). Remaining steps for this issue: render cut → mutation cut → remove the `appStore` `monitors`/`monitoringStatsCache` reducers.

armaxri:
Mutation cut landed in #2249 (merged into develop).

[...]

Issue #2153, opened 2026-07-30
Phase 5: the long tail — migrate remaining domains one at a time


armaxri:
> ⚠ **BLOCKED on the projection-transport decision — #2139 Open Design Decision #1, "Projection granularity & transport". DO NOT START until the maintainer settles it,** and until Phase 1 · Substrate (#2149) has landed.

## Scope

**Phase 5 · The long tail.** Migrate the remaining domains onto the Rust store + projector, one at a time (one PR each), so `appStore.ts` shrinks monotonically. Remaining domains per the concept:

- Connections tree
- File browsers
- Transfers (transfer queue state machine)
- Agents
- Monitors
- Settings

Each domain is a self-contained migration: authoritative state → Rust store, projected view model, mutations → intents, delete the TS reducers. **File a dedicated sub-issue per domain when Phase 1 lands** (this issue is the umbrella) — do not attempt them as one PR.

## Acceptance criteria (umbrella)

- Each listed domain migrated in its own PR; `appStore.ts` authoritative state + reducers for that domain removed.
- Projection-assertion tests cover each migrated domain.
- The store's authoritative footprint trends to ~zero as domains land.

## References

- Concept: `docs/concepts/future/stateless-ui-agent-centric.html` — Phase 5 · The long tail; "What happens to the 8.3k-line appStore".
- Part of #2139.

## Dependency / blocker note

**Blocked on Open Design Decision #1 (projection transport).** Depends on **Phase 1 · Substrate (#2149)**. Sequence domains one at a time; the plugin-frontend runtime is the hard exception (Open Design Decision #3) and is out of scope here.

armaxri:
**Unblocked.** The ⚠ BLOCKED note is stale: Open Design Decision #1 (projection transport) is settled (multi-client → per-region versioned diff channels), the Phase-1 substrate merged (#2163), and Phases 3 (layout) + 4 (session) have proven the per-domain shadow→render-cut→mutation-cut pattern. Under the settled **phased/per-domain** decision each long-tail domain migrates independently, so Phase 5 does NOT wait on Phase 4's tail. Decomposing into per-domain sub-issues now and grinding them one at a time.

armaxri:
## Phase 5 decomposed into per-domain sub-issues

[...]

PR #2260, opened 2026-07-31
feat(backend): shadow settings projection region (#2227)

armaxri:
## Summary

Backend-only **shadow foundation** for the **Settings** domain — Phase 5 of the stateless-UI migration. Adds `src-tauri/src/settings_projection/`: a `SettingsStore` modeling the `appStore` `AppSettings` document, exposed as a single **shared** `settings` projection region plus `settings.*` intents, mirroring the recent shared shadows (system-monitor #2224, agents #2226).

**Zero user-facing change.** Nothing in the live UI subscribes to or renders the region, and no frontend code dispatches `settings.*` intents. `appStore`'s `settings` slice and its `updateSettings` / `saveSettings` persistence remain authoritative. The diff touches only `src-tauri/` — no `src/` changes.

## What the shadow models

The app-settings slice the frontend drives (`appStore` `settings` / `savedSettings` — the persisted `AppSettings` document). `AppSettings` is one large, open-ended, forward-compatible JSON document the app loads and saves as a whole, and the frontend shape carries presentation keys the backend's typed struct doesn't mirror. So — like the backend already treats `customThemes` / `syntaxHighlighting` as opaque `serde_json::Value` — the store models the document **opaquely** (a JSON object), owning the whole-document replace + shallow-patch semantics without duplicating (and drifting from) the ~50-field typed struct. This is the coherent core; per-key typing is neither required by the later render/mutation cuts nor desirable.

### Intents

| kind | payload | effect |
| --- | --- | --- |
| `settings.replace` | `{ settings: {…} }` | overwrite the whole document (mirrors `updateSettings` / `saveSettings`) |
| `settings.patch` | `{ patch: {…} }` | shallow-merge top-level keys (mirrors the `{ ...current.settings, <field> }` spreads, e.g. `updateShellIntegration`) |
| `settings.reset` | `{}` | reset to the default baseline |

## Region scope — Shared (Open Design Decision #4 / #6)

`AppSettings` is a **single persisted document** every client sees: in a multi-client world every client sees the same settings and an edit projects to all — like SSH tunnels and the shared session-lifecycle region. So a single **shared** `settings` region, seeded at version 0 with the frontend default document. Any device-local preference that should *not* sync (were one ever carved out) would become a client-scoped presentation region instead — none exists today, so the whole document is shared here.

## Tests

- **Store unit tests** — default snapshot, whole-document replace, shallow patch (insert / overwrite / null-set / untouched keys), reset-to-defaults.
- **Projection-contract tests** (over the #2164 harness) — subscribe snapshot identical to every subscriber, one coalesced diff fanned to two subscribers with monotonic versions, rejection paths advance nothing, a no-op advances nothing, a dead subscriber is reaped, and the client cache converges across a replace → patch → reset lifecycle.

`cargo test -p termihub settings_projection` (12 pass), `cargo clippy -p termihub --all-targets --all-features -- -D warnings`, and `cargo fmt --all --check` all green locally.

Part of #2227 · Part of #2153 · Part of #2139

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -65,6 +65,13 @@ mod session_history;
 /// the ported auto-reconnect engine (#2144). Registered and served but not yet
 /// driving the live UI — see [`session_projection`].
 mod session_projection;
+/// Shadow app-settings authority (#2227, Phase 5 of #2139/#2153): the shared
+/// `settings` projection region + `settings.*` intents modeling the `appStore`
+/// `AppSettings` document (the persisted user-preferences slice behind
+/// `updateSettings` / `saveSettings`), held as an opaque JSON document.
+/// Registered and served but not yet driving the live UI — see
+/// [`settings_projection`].
+mod settings_projection;
 mod spawn;
 /// Shadow system-monitor authority (#2224, Phase 5 of #2139): the shared
 /// `system-monitors` projection region + `monitor.*` intents, built on the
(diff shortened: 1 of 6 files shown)


armaxri:
Follow-up filed from this work: #2261 — the backend `AppSettings` struct silently drops several persisted frontend settings (`confirmCloseAttachedTab`, `warnLargePortScan`, `warnLargePingSweep`, `screenReaderMode`) on save, so their off-state doesn't survive a restart. Out of scope for this backend-only shadow.

Issue #2251, opened 2026-07-31
serve plugin code via a custom Tauri protocol to drop blob: from script-src (sandbox last-mile)


armaxri:
Follow-up to #2136 (PR #2250). PR #2250 moved frontend plugin execution into a least-privilege Web Worker sandbox, but plugin code still enters the worker via a `blob:` URL (`importScripts` of a `Blob`), so `script-src` must retain `blob:`:

```
script-src 'self' blob: 'wasm-unsafe-eval'
```

This is the **final** step of the original #2136 scope: remove `blob:` from `script-src` once plugin code no longer needs it to execute.

## Context
- Plugin JS is now confined to the sandbox worker (`src/plugins/sandbox/pluginSandboxWorker.ts`), which does `URL.createObjectURL(new Blob([wrapped]))` + `importScripts(url)`. That `importScripts` of a blob URL is governed by `script-src` (not `worker-src`), which is why `blob:` cannot be dropped yet.
- `worker-src 'self' blob:` and `child-src 'self' blob:` are a separate allowance (kept — Monaco/xterm workers) and are NOT affected by this issue.

## Scope
- [ ] Load plugin code into the worker without a `blob:` script — e.g. fetch/importScripts from a plugin-served `'self'` origin (a Tauri custom-protocol asset URL for `<config>/plugins/<id>/<entry>`), or another non-`blob:` mechanism that still runs under a `script-src` without `blob:`.
- [ ] Remove `blob:` from `script-src` in `src-tauri/tauri.conf.json` and mirror in `src-tauri/tauri.test.conf.json` (the `cspConfig.test.ts` drift-lock requires them to match except `connect-src`).
- [ ] Extend `src/security/cspConfig.test.ts` to assert `script-src` no longer contains `blob:`.
- [ ] Verify the plugin sandbox (parser + status-bar widget) still loads and works under the tightened CSP in a real build via `tests/system/tests/test_plugin_sandbox.py` (integration lane — the enforced CSP is only present in a production-shaped build).

## Done when
- Frontend plugin code executes with no `blob:` in `script-src`.
- The production `script-src` no longer contains `blob:`.
- `test_plugin_sandbox.py` still passes (widget renders + parser transforms + zero CSP violations) under the tightened CSP.

armaxri:
## Investigation: dropping `blob:` from `script-src` requires a backend loader substrate, not a config tweak

[...]

PR #2265, opened 2026-07-31
feat(backend): add plugin:// uri-scheme protocol serving plugin files (Part of #2251)

armaxri:
## What

Adds a Tauri custom URI-scheme protocol (`plugin_protocol.rs`) that serves files from an installed plugin's directory (`<app-data>/plugins/<id>/<path>`) over an **app-controlled origin**:

- macOS / iOS / Linux: `plugin://localhost/<id>/<path>`
- Windows / Android: `http://plugin.localhost/<id>/<path>`

Registered on the Tauri builder via `register_uri_scheme_protocol`.

## Why (Part of #2251 / #2136)

The frontend-plugin sandbox worker currently runs plugin code by turning its source into a `blob:` URL and `importScripts`-ing it, which is the only remaining `script-src blob:` consumer. To drop `blob:` from `script-src`, plugin code must load from a URL whose origin already satisfies `script-src`. This protocol is that origin — the **substrate** the loader will `importScripts` from instead of a blob.

This PR is the **first slice** of #2251. It deliberately does **not** change the CSP or the loader, so there is zero CSP-regression surface here: the scheme is registered but not listed in `script-src`, so the webview cannot yet reach it (the handler is exercised entirely by unit tests). The loader switch + CSP change + `blob:` removal + cross-platform CSP verification is the next slice — see the linked follow-up.

## Path safety

Every read delegates to `PluginManager::read_file`, reusing its already-tested plugin-id slug validation and `..`/absolute traversal refusal (single source of truth for containment). The handler adds percent-decoding + request-shape validation and **fails closed**: any malformed request or read error returns an empty `400`/`403`/`404`/`500`, never a partial or misattributed body.

## Tests

10 new Rust unit tests over `serve()` and the parse/content-type helpers: id+relative split, percent-decoding, JS/JSON content-type mapping, traversal rejection (relative and id), missing plugin/file → 404, malformed request → 400. `cargo clippy --all-targets -D warnings` clean; `cargo fmt` clean. Frontend CSP + sandbox guards unchanged and green (60 tests).

## Verification note

No production-CSP plugin run is applicable to this slice: it changes no CSP directive and no plugin loading/runtime path (the worker still uses `blob:`). That runtime acceptance gate — enable a plugin, confirm its widget/parser runs in the worker under the new CSP with zero violations, cross-platform — belongs to the loader-switch slice, where it is mandatory.

`blob:` is **not** yet removed from `script-src`; #2251 and #2136 stay open.

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -58,6 +58,9 @@ tauri-plugin-fs = "2"
 notify-debouncer-full = "0.7"
 serde = { workspace = true }
 hex = "0.4"
+# Percent-decoding paths from the `plugin://` custom URI-scheme protocol (#2251),
+# so encoded characters in a request path resolve to the real on-disk filename.
+percent-encoding = "2"
 # SHA-256 hashing: agent-binary integrity verification (all platforms) and
 # Windows VcXsrv checksum validation.
 sha2 = "0.10"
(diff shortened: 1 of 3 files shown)


armaxri:
Next slice tracked in #2266 (loader switch + `blob:` removal + cross-platform CSP verification).

Issue #2387, opened 2026-08-02
Phase-5 step-1: make the Transfers projection store server-authoritative (prereq for #2229)


armaxri:
**Part of the thin-UI finalize (#2139 / #2153).** Server-authority step-1 prerequisite for the **Transfers** domain reducer-removal (#2229). Per the #2139 go-forward plan, the projection regions are client-side mirrors seeded from `appStore` and never fed from the server-side data source, so #2229's reducer-removal would be a lossy inversion unless this lands first. This is the **closest analog to Monitors #2376 (PR #2379)**: a backend engine loop produces authoritative progress and emits it straight to the frontend while the store only sees client re-dispatch.

Audit (2026-08-02): `TransferStore` is written **only** by frontend `transfer.*` intents; no server writer. Backend progress is produced in `files/transfer/` and emitted as the `TRANSFER_PROGRESS_EVENT` Tauri event; lifecycle transitions live in the registry/scheduler.

## Scope (medium)
- [ ] Fold each registry/scheduler lifecycle transition + progress sample into `TransferStore` **at the source** (register / queue / progress / pause / resume / finish / cancel), mirroring the `monitoring_controller` server-side push-task template from #2376.
- [ ] Additive only — no UI change, no `appStore.ts` change.

## Server files
`src-tauri/src/files/transfer/registry.rs` (`TransferRegistry`, `TransferHandle::transition`), `src-tauri/src/files/transfer/mod.rs` (`app_progress_sink` / `TRANSFER_PROGRESS_EVENT`), `src-tauri/src/files/transfer/scheduler.rs`, `src-tauri/src/commands/transfer.rs`.

## Sequencing
Disjoint from Agents / File-browsers / Connections / Settings step-1s — **can run in parallel** with any of them.

## Done when
`TransferStore` is fed by the backend transfer engine at the source, making #2229's subsequent reducer-removal non-lossy.

Issue #2376, opened 2026-08-01
prerequisite for #2224: fold monitoring stats/status into SystemMonitorStore server-side


armaxri:
Prerequisite for the final step of #2224 (Phase 5 Monitors migration — the reducer removal / data-flow inversion). Discovered while implementing #2224.

## The blocker

The reducer-removal step of #2224 requires making the `system-monitors` projection region **authoritative** so the appStore `monitors` / `monitoringStatsCache` reducers can be deleted. That cannot be done safely today because **the `SystemMonitorStore` is never fed from the data source** — it is a pure client-driven mirror:

- The authoritative monitoring data (stats + status) is produced **server-side** by the SSH/session collector loop and emitted straight to the frontend as Tauri events in `src-tauri/src/session/monitoring_controller.rs` (`session-monitoring-stats` / `session-monitoring-status`). **This code never touches `SystemMonitorStore`.**
- `SystemMonitorStore` (managed in `src-tauri/src/lib.rs` `setup()`) is updated **only** by the client-dispatched `monitor.*` intents routed in `src-tauri/src/system_monitor_projection/projection.rs`.
- So the current flow for a stats sample is: backend collector → Tauri event → **frontend** folds into appStore + best-effort `mirrorMonitorIntent("monitor.stats")` → store → diff → frontend render.

Removing the appStore reducers (the render/mutation fallback) makes the store authoritative, which means **every** monitoring sample would reach the UI only via a client round-trip (`event → client dispatch → store → diff → render`) **with no local fallback**. A dropped/rejected dispatch or a lost diff silently drops a safety-critical monitoring sample — there is no server-side source of truth to recover it from, because the backend never folded it into the store. This is a lossy inversion and is unacceptable for ventilator-grade monitoring. It is most acute in remote-client mode (the sample round-trips over the link back to the store) but is a correctness change even in-process on desktop (a "cannot fail" local fold becomes a "can fail" async round-trip).

## The fix (make the store authoritative at the source)

Fold monitoring transitions into `SystemMonitorStore` **server-side**, where the data originates, so the store is genuinely authoritative and the frontend render cut simply reads it (no client re-dispatch of stats/status, no round-trip, no loss):

[...]

Issue #2225, opened 2026-07-30
Phase 5: migrate the Connections tree domain to a Rust store + projection region


armaxri:
Part of #2153 (Phase 5 · the long tail) and #2139 (stateless-UI port).

Migrate the **Connections tree** domain (the saved-connection / folder tree the sidebar renders) onto a Rust store + projection region, following the proven per-domain pattern (Phases 3 & 4).

### Region scope (Decision #4)
Likely **shared** — saved connections/folders are a single persisted config shared by all clients (like tunnels). Confirm at shadow time: any per-client expand/collapse or selection state is presentation and stays a frontend concern under partial projection (as with layout).

### Scope
- [ ] **Shadow**: `src-tauri/src/connections_projection/` — a store modeling the authoritative connection/folder tree + a `connections` region + `connection.*` intents (create/edit/remove/move/reorder folder+entry), UI NOT cut over. Projection-assertion tests (#2164 harness). Zero user-facing change.
- [ ] **Render cut**: the connection-tree sidebar reads the projected `connections` region (parity-safe).
- [ ] **Mutation cut**: flag-gated — tree mutations dispatch `connection.*` intents.
- [ ] **Remove reducers**: delete the connection-tree appStore state + reducers once the region is authoritative.

### Done when
`appStore.ts` no longer holds connection-tree authoritative state; the sidebar renders from the `connections` projection; projection-assertion tests cover the domain; nothing user-facing changed at each step.

### Sequencing
Independent of Monitors/Agents/Settings. The saved-connection config already lives in Rust (`connection/`), so the store wraps existing authority rather than re-homing it.

armaxri:
Coordinator triage (2026-07-31): verified against develop — this domain's **shadow, render-cut (`useProjected*` hook), and mutation-cut (`*IntentsEnabled`, on by default) are all merged.** The only remaining checkbox is the appStore reducer-removal, which per the domain-migration pattern is **deferred until after the extended-testing period** ("once the region is confirmed authoritative in the wild"). Removing `Ready2Implement` so it isn't picked up as ready work (two agents already burned tokens walking into already-done Phase-5 issues today); re-add it when the deferred removal is cleared to proceed.

armaxri:
[...]

PR #2393, opened 2026-08-02
feat(connection): feed the connections projection store server-side (Closes #2389)

### LINKED ISSUE #2389: Phase-5 step-1: make the Connections projection store server-authoritative (prereq for #2225)
**Part of the thin-UI finalize (#2139 / #2153).** Server-authority step-1 prerequisite for the **Connections** domain reducer-removal (#2225). Per the #2139 go-forward plan, the projection regions are client-side mirrors seeded from `appStore` and never fed from the server-side data source, so #2225's reducer-removal would be a lossy inversion unless this lands first. Mirrors **Monitors #2376 (PR #2379)**.

Audit (2026-08-02): `ConnectionsStore` is written **only** by frontend `connection.*` intents. Its docstring claims it "wraps the existing Rust authority," but that is only **type reuse** — it is a separate object seeded empty in `lib.rs` and never read from the real authority (`ConnectionManager` + persisted config).

## Scope (small–medium — straightforward, the manager already holds the exact wire types)
- [ ] Fold the `ConnectionManager` saved-connection/folder mutations (add / update / remove / move / addFolder / removeFolder / toggleFolder-persist) into `ConnectionsStore` **at the source**.
- [ ] Additive only — no UI change, no `appStore.ts` change.

## Server files
`src-tauri/src/commands/connection.rs`, `src-tauri/src/connection/manager.rs`, possibly `src-tauri/src/connection/storage.rs`.

## Sequencing
**Collides with the Settings step-1 (#2227 prereq)** — both touch `commands/connection.rs` + `connection/manager.rs`. Do NOT run these two in parallel; sequence them. Disjoint from Agents / File-browsers / Transfers.

## Done when
`ConnectionsStore` is fed by the `ConnectionManager` at the source, making #2225's subsequent reducer-removal non-lossy.

armaxri:
## What

Makes the shared \`ConnectionsStore\` (the \`connections\` projection region) **server-authoritative at the source**: every backend-produced saved-connection / folder mutation is reflected into the store from the persisted \`ConnectionManager\` authority the instant it lands, instead of relying solely on client-dispatched \`connection.*\` intents.

Prerequisite for #2225 (Phase 5 Connections reducer-removal). Audit (from #2389): \`ConnectionsStore\` was written **only** by frontend \`connection.*\` intents and seeded empty in \`lib.rs\` — its docstring's "wraps the existing Rust authority" was only *type reuse*, never fed from the real authority (\`ConnectionManager\` + persisted config). Removing the appStore reducers (making the store authoritative) would be a **lossy** inversion while the store is never fed from the data source. This PR feeds it from the source so #2225 can invert the data flow with no data-loss risk. Closest analogs: Monitors #2376 (PR #2379), Transfers #2387 (PR #2391).

## Changes

- **New \`fold_connections_from_manager\`** (\`connections_projection/projection.rs\`): resolves the managed \`ConnectionsStore\` + \`ConnectionManager\` + \`ProjectionState\` from the \`AppHandle\`, reflects the manager's **whole authoritative post-mutation tree** into the store (\`get_all\` → \`replace\`), and publishes the \`connections\` region diff. Best-effort (skips if any of the three is unmanaged, or the disk reload fails); never holds the store lock across an await.
- **Fold at the source** (\`commands/connection.rs\`): each command that mutates the persisted saved-connection / folder tree — \`save_connection\`, \`delete_connection\`, \`move_connection_to_file\`, \`save_folder\` (also the \`toggleFolder\`-persist edge), \`delete_folder\`, \`import_connections\`, \`import_connections_with_credentials\` — folds after the manager mutation succeeds. (\`AppHandle\` is Tauri-injected — no change to the JS invoke surface.)
- **Startup seed** (\`lib.rs\`): the \`connections\` region is now seeded from the persisted manager tree instead of an empty baseline, so a subscriber attaching before the first mutation sees the real saved-connection tree.
- **Tests** (\`projection_tests.rs\`): drive the production fold end to end against \`tauri::test::mock_app()\` with a real \`ConnectionManager\` — a persisted mutation feeds the store and fans one region diff out **with no client dispatch**; the store reflects the manager's **recomputed** path-based id (not the caller's optimistic id); a folder/connection lifecycle mirrors the manager authority (delete-folder re-homes the child to root); a no-op when unmanaged.

## Why reflect the whole tree, not per-op folds

Unlike Transfers/Monitors (which fold a single event via one store method), the \`ConnectionManager\` is a **coarse** authority: one \`save_connection\` may recompute the path-based id, deduplicate sibling names, re-home children, and migrate credentials. Replaying the intent-level ops (\`add\`/\`update\`/…) against the store would drift from the persisted truth, so the fold reflects the manager's authoritative snapshot (\`get_all\` → \`replace\`). The projector coalesces an unchanged snapshot to no diff, so a mutation that leaves the main tree untouched (e.g. an external-file-only save) is a no-op.

## Additive / no user-facing change

- The per-transition \`connection.*\` intents and the render-cut \`connection.replace\` mirror stay in place; \`appStore.ts\` and the frontend mirror are intentionally untouched.
- Nothing in the live UI subscribes to the \`connections\` region yet. The render/mutation inversion (dropping the redundant client re-dispatch, making the store authoritative) is the **later #2225 step**.
- Backend-only, no user-facing behavior change → no changelog fragment.

## Scope note (for #2225)

The store reflects the **main** persisted store (the tree the listed manager mutations affect). Connections merged in from **external** connection files are a read-only overlay not modeled here — filed as a follow-up so #2225's render cut can decide whether the region needs them.

## Verification

- \`cargo test --lib connections_projection\` — 29 pass (3 new)
- \`cargo test --lib connection::\` — 195 pass
- \`cargo build\`, \`cargo fmt --check\`, \`cargo clippy --all-targets --all-features\` — clean

Closes #2389

### DIFF
--- a/src-tauri/src/commands/connection.rs
+++ b/src-tauri/src/commands/connection.rs
@@ -1,7 +1,7 @@
 use std::sync::{Arc, Mutex};
 
 use serde::Serialize;
-use tauri::State;
+use tauri::{AppHandle, State};
 use tracing::{debug, info};
 
 use crate::connection::config::{
@@ -68,12 +68,17 @@ pub fn load_connections_and_folders(
 #[tauri::command]
 pub fn save_connection(
     connection: SavedConnection,
+    app: AppHandle,
     manager: State<'_, ConnectionManager>,
 ) -> Result<String, String> {
     debug!(id = %connection.id, name = %connection.name, "Saving connection");
-    manager
+    let persisted_id = manager
         .save_connection_routed(connection)
-        .map_err(|e| e.to_string())
+        .map_err(|e| e.to_string())?;
+    // Server-authority fold (#2389): reflect the persisted tree into the shadow
+    // `ConnectionsStore` at the source. Additive; no user-facing change.
+    crate::connections_projection::projection::fold_connections_from_manager(&app);
+    Ok(persisted_id)
 }
 
 /// Delete a connection by ID, optionally from an external file.
@@ -81,12 +86,15 @@ pub fn save_connection(
 pub fn delete_connection(
     id: String,
     source_file: Option<String>,
+    app: AppHandle,
     manager: State<'_, ConnectionManager>,
 ) -> Result<(), String> {
     info!(id, ?source_file, "Deleting connection");
     manager
         .delete_connection_routed(&id, source_file.as_deref())
-        .map_err(|e| e.to_string())
+        .map_err(|e| e.to_string())?;
+    crate::connections_projection::projection::fold_connections_from_manager(&app);
+    Ok(())
 }
 
 /// Move a connection between storage files (main <-> external).
@@ -95,6 +103,7 @@ pub fn move_connection_to_file(
     connection_id: String,
     current_source: Option<String>,
     target_source: Option<String>,
+    app: AppHandle,
     manager: State<'_, ConnectionManager>,
 ) -> Result<SavedConnection, String> {
     info!(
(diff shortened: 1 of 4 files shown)


armaxri:
Follow-up filed for the external-file-connections scope note above: #2394 (reflect external-source connections in the server-authoritative `connections` region).

Issue #2241, opened 2026-07-31
Phase 4 step 5a: cut the restore-cohort machine over to the server-side authority


armaxri:
Follow-up to #2206 (PR #2240). The shadow `RestoreCohortStore` + `restore-cohort@<clientId>` region + `restore.*` intents landed in PR #2240; this cuts the live UI over to it, following the proven session-lifecycle sequence (shadow → render-cut → flag-gated mutation-cut → keep-reducers-as-fallback).

## Scope
- [ ] **Render-cut:** subscribe the frontend to the client-scoped `restore-cohort@<clientId>` region; render the aggregate summary toast from the projected `settlement` (fire once per new monotonic `seq`) and the bulk "Reconnect failed tabs" action from the projected `failedTabIds`. Apply the **live-terminal filter** frontend-side (intersect `failedTabIds` / `settlement.retryTabIds` with live terminal tabs) — that stays a frontend concern because it needs the tab registry.
- [ ] **Mutation-cut (flag-gated):** route `beginRestoreCohort` / `settleRestoreTab` through `restore.beginCohort` / `restore.settleTab`, behind a `localStorage` flag mirroring `termihub.sessionIntents`, with a parity fallback.
- [ ] Keep the `appStore` cohort reducers as the resilience/rollback fallback (per the #2205 reframe), not deleted.
- [ ] Parity tests for the cut; confirm no user-visible change to restore/launch feedback.

## Done when
The restore-cohort summary + bulk-retry render from the projected region and the cohort transitions dispatch `restore.*` intents (flag-gated with fallback); parity tests green; no user-visible change.

Issue #1227, opened 2026-07-07
Add bulk "reconnect failed tabs" control after a partial restore (audit M2)


armaxri:
Follow-up to #1146 (PR #1226).

PR #1226 added the aggregate partial-restore **summary** toast (audit GAP G4): after a restore/launch fans out, one toast reports "Restored N of M tabs — K could not reconnect". The summary is informational only — recovery is still per-tab (click into each failed tab and Reconnect).

This issue covers the remaining **M2** control from the workspace save/restore audit: a single "Reconnect failed tabs" action so a user can retry every failed tab from the partial restore at once.

### Scope
- [ ] Surface a bulk "Reconnect failed tabs" affordance after a partial restore (e.g. an action on the summary toast, or a control in the Workspace sidebar / status bar).
- [ ] Drive every failed tab of the settled cohort through the existing per-tab reconnect path (`reconnectTerminal` / `retryTerminalSpawn`) in one action.
- [ ] Give feedback on the bulk retry (pending → success/partial), reusing the cohort summary if practical.
- [ ] Tests covering: bulk retry re-drives only the failed tabs; a subsequent partial outcome re-summarizes correctly.

### Done when
A user who gets "Restored N of M tabs — K could not reconnect" can retry all K failed tabs with a single control, with feedback, instead of reconnecting each tab individually.

Reference: `docs/audits/workspace-save-restore-state-machine.md` (M2 in §6, and G4).

Issue #1146, opened 2026-07-06
Implement workspace save/restore state-machine fixes (from #1135 audit)


armaxri:
Implementation follow-up for the **workspace save/restore** state-machine audit.

- **Audit doc:** [`docs/audits/workspace-save-restore-state-machine.md`](../blob/develop/docs/audits/workspace-save-restore-state-machine.md)
- **Audit issue:** #1135
- Each gap cites `file:line`, symptom, and smallest fix in the audit doc.

The audit recommends an explicit `workspaceRestore` status field (`status`/`total`/`failed`/`launchingId`) that most gaps collapse into.

### High (leak / data-loss)

- [ ] **G1** — tear down existing live sessions before `launchWorkspace`/`restoreLastSession` place new tab groups (today prior PTY/SSH/agent sessions are orphaned into Open Connections).
- [x] **G2** — `await` the save with try/catch + toast; keep the dialog open on error (today a failed save closes the dialog and looks successful). _(done — #1158)_
- [ ] **G3** — surface `LoadFailed`/`EmptyBuild`/corrupt-file restore failures (toast + a recovery channel for `last-session.json`, like `workspaces.json` already has).

### Medium

- [ ] **G4** — track the restored-tab cohort and raise a summary ("Restored N of M, K failed"); add a bulk "reconnect failed tabs".
- [ ] **G5** — gate auto-save behind a `restoreInProgress` flag so a mid-restore snapshot doesn't overwrite the good session file.
- [ ] **G6** — disable the Launch control + guard re-entry while a launch is in flight.

### Low

- [x] **G7** — confirm/undo + toast on workspace delete; only mutate local state after the backend resolves. _(done — #1179)_
- [x] **G8** — surface import/export results (imported count) and errors instead of empty catches. _(done — #1183)_

### Missing controls

- [ ] Launch-will-close-current confirm, Retry-all, manual "restore last session now", save feedback, restore-failed notice, cancel-during-launch, delete confirm/undo.

Follow the TDD workflow and split into separate PRs per logical topic as appropriate.

armaxri:
All mergeable gaps from the workspace save/restore audit are done: G2 (#1158), G3 (#1205), G5 (#1212), G6 (#1198), G7 (#1179), G8 (#1183), G1 teardown (#1216), G4 partial-restore summary (#1226). Bulk 'reconnect failed tabs' (missing-control M2) split to follow-up #1227. Closing this implementation umbrella.

PR #2435, opened 2026-08-07
refactor(workspace): make the restore-cohort region authoritative and remove the appStore cohort reducers (Part of #2206)

armaxri:
## What

Phase 4 step 5 of #2139: the **restore-cohort** domain's reducer removal — the direct analog of the monitors (#2385) and transfers (#2399) removals. Since the render + mutation cut (#2241) was live, this completes the inversion: the client-scoped `restore-cohort@<clientId>` region (`RestoreCohortStore`) is now the **sole source of truth** for the aggregate restore/launch feedback (#1146 / #1227).

A **contained appStore-internal data-flow inversion**: the domain has **0 component readers** (the cohort is client-owned orchestration that only fires the aggregate restore-summary toast), so there is no component migration.

## Changes

- **`appStore`**: removed the `restoreCohort` + `failedRestoreTabIds` slice and the `settleRestoreCohort` reducer.
  - `beginRestoreCohort` / `settleRestoreTab` are now thin dispatchers of the `restore.beginCohort` / `restore.settleTab` intents; the region is the **sole guard** that ignores a stray/duplicate settle (the frontend no longer second-guesses).
  - The aggregate summary toast fires from the **projected settlement** diff via a renderer the store registers with the bridge — the store owns it because firing the toast and intersecting the retry set with the **live terminal tabs** need the tab registry (deliberately frontend-side).
  - `reconnectFailedRestoreTabs` reads the captured failed-tab set from `currentRestoreCohortView()`.
- **`restoreCohortBridge`**: removed the render/mutation feature flags + the local-fallback machinery (matching monitors/transfers going authoritative); added `currentRestoreCohortView()` and the settlement renderer registration; `onRegionChange` fires the renderer once per new monotonic settlement `seq`.
- **Rust** (`store.rs`, `lib.rs`): doc comments updated from "shadow / not authoritative" to authoritative. No behavior change (the store was already complete).
- **Tests**: extracted a shared `restoreCohortHarness` (in-memory region twin), and reworked the cut / summary / bulk-reconnect suites to drive the cohort through the authoritative region (async settlement) instead of the removed local slice.

## Non-lossiness

The cohort is client-owned (no server feed to miss). Verified the cohort still forms, settles, and fires its summary toast correctly with the reducer gone — the settlement signal still comes from the client-dispatched session connect/exit events. A rejected dispatch fires no toast and logs (the authoritative model has no local fallback), consistent with monitors/transfers.

## Testing

- Full frontend suite: **4873 passed**. Restore-cohort suites (bridge, mutation, render, summary, bulk-reconnect): 20 passed.
- Rust `restore_cohort_projection`: 20 passed.
- `tsc --noEmit`, ESLint, Prettier, `cargo fmt --check` clean.

Part of #2206

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -57,12 +57,12 @@ mod plugin_protocol;
 /// versioned diff channels with multi-subscriber fan-out. Public so integration
 /// tests can drive the projector directly.
 pub mod projection;
-/// Shadow restore-cohort authority (#2206, Phase 4 step 5 of #2139): the
-/// client-scoped `restore-cohort@<clientId>` projection region + `restore.*`
+/// Authoritative restore-cohort state machine (#2206, Phase 4 step 5 of #2139):
+/// the client-scoped `restore-cohort@<clientId>` projection region + `restore.*`
 /// intents, modelling the startup restore/launch cohort aggregation (#1146 /
 /// #1227) on top of the ported restore-decision engine (`termihub_core::restore_mode`,
-/// #2145). Registered and served but not yet driving the live UI — see
-/// [`restore_cohort_projection`].
+/// #2145). The sole source of truth driving the live UI's aggregate summary toast
+/// (reducer removal) — see [`restore_cohort_projection`].
 mod restore_cohort_projection;
 pub mod run_location;
 mod session;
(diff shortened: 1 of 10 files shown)


armaxri:
Superseded by #2436 — recommitted with a commitlint-compliant subject (the original 105-char header exceeded the 100-char limit). Identical diff, no force-push.

Issue #2205, opened 2026-07-30
Phase 4 step 4: remove authoritative session-lifecycle state from appStore


armaxri:
Follow-up to #2152 (PR #2202). Phase 4 step 4 of #2139. **Blocked on step 3** — do not label `Ready2Implement` until step 3 lands.

## Scope
- [ ] Delete the authoritative lifecycle state + reducers from `appStore.ts`: `terminalConnecting`, `terminalDisconnects`/disconnect reasons + errors, `terminalReconnectingTabs`, `terminalReconnectTriggerErrors`, `terminalAutoReconnect`, and the module-scope `autoReconnectTimers`.
- [ ] Keep only per-client *presentation* affordances that are genuinely not lifecycle (dismissed disconnect overlay / "view mode" scrollback browsing).

## Done when
- `appStore` holds no authoritative session-lifecycle state; the store shrinks; the region is the single source of truth.
- No user-visible change; existing lifecycle tests pass against the projected source.

Blocked on step 3. Blocks step 5.

armaxri:
Coordinator triage (2026-07-31): verified against develop — the server-side move + mutation-cut for this step's domains are merged (`*IntentsEnabled` flags present, on by default). The remaining work is the appStore authoritative-state removal, **deferred until after the extended-testing period** per the domain-migration pattern. Removing `Ready2Implement` so it isn't picked up prematurely; re-add when the deferred removal is cleared.

armaxri:
Re-adding `Ready2Implement`: the maintainer un-deferred reducer-removal (2026-08-01), so this domain's final step — remove the appStore authoritative slice now that its projection is authoritative — is actionable again (proceed-on-parity-tests). The label was stripped during triage only because removal was deferred at the time.

armaxri:
[...]

Issue #2203, opened 2026-07-30
Phase 4 step 2: cut session-lifecycle mutations to session.* intents (+ backend timer driver)


armaxri:
Follow-up to #2152 (PR #2202). Phase 4 step 2 of #2139. The shadow `SessionLifecycleStore` + shared `session-lifecycle` region + `session.*` intents landed in PR #2202 (not yet driving the UI). This step makes the store **authoritative for session status** while rendering still reads `appStore` (the bridge reconciles projected diffs back), mirroring Phase-3 layout step 2 (#2188).

## Scope
- [ ] Frontend bridge: on session create, seed the store's status for the new session (seed-before-mutate), then dispatch `session.*` intents for the transitions the frontend currently drives locally — `connect`, `connected`, `connectFailed`, `disconnect`, `dropped`, `reconnect`, `reconnectAttempt`, `reconnectFailed`, `cancelReconnect`, `remove`.
- [ ] **Backend timer driver**: the auto-reconnect loop's `setTimeout` currently lives in `appStore` (`autoReconnectTimers`). Once the loop is server-authoritative, the backend must arm the `delayMs` a `Waiting` state reports and fire `session.reconnectAttempt` itself. Add this driver to `session_projection` (tokio timer keyed by session id; cancel on cancel/remove/success).
- [ ] Keep `appStore` reducers as the render source for this step; reconcile projected diffs into them (bridge), so there is still zero user-facing change.

## Done when
- The frontend dispatches `session.*` intents for all lifecycle transitions and the backend store is authoritative for status + reconnect loop (incl. the timer).
- No user-visible change to connect/reconnect/disconnect behaviour.
- Projection + bridge tests cover the seed-then-mutate round trip.

Blocks: step 3 (render cut).

PR #2438, opened 2026-08-07
feat(session): server-authoritative session-lifecycle identity bridge + initial-connect fold (Closes #2431)

### LINKED ISSUE #2431: Phase-4 step-1: make the session-lifecycle projection store server-authoritative (prereq for #2205)
Phase-4 twin of the Phase-5 server-authority step-1s (#2376 Monitors, #2389 Connections). Per the Phase-4 audit (2026-08-07): the shared `session-lifecycle` projection store (`src-tauri/src/session_projection/`) is currently a **client mirror** — written only by the `session.*` intent routes (`session_projection/projection.rs:81`), seeded empty in `lib.rs`. A genuine **server-produced transition stream bypasses it**, so the #2205 reducer-removal would be a lossy inversion unless this lands first.

## Server-produced transitions to fold at the source
Add a `fold_session_transition(app_handle, …)` (twin of `fold_monitor_transition`) and call it where these originate, publishing the shared `session-lifecycle` region diff:
1. **Unexpected drop:** `terminal-exit` emitted at `src-tauri/src/session/manager.rs:158` (`emit_exit`) → `store.dropped(id, err)`.
2. **Connect outcome:** the session connect command (`commands/session.rs`) → `store.connect` / `store.connected` / `store.connect_failed`.
3. **Graceful disconnect / cancel:** the disconnect/cancel command → `store.disconnect` / `store.remove`.

The `SessionLifecycleStore` transition methods already exist + are unit-tested; this only wires them to the server events. The reconnect backoff timer is already server-side (`ReconnectTimerDriver`, #2203) — only the entry transitions remain. Additive, server-side, no UI/appStore change.

## Server files (collision map)
`src-tauri/src/session/manager.rs` (emit_exit — unique to this domain), `src-tauri/src/commands/session.rs` (connect/disconnect/cancel outcomes; also touched by the merged Monitors #2379 — no live collision), possibly `session/remote_proxy.rs`/`session/graphical_manager.rs` for agent/remote+graphical sessions.

## Done when
The shared `session-lifecycle` region reflects drop/connect/connected/connect_failed/disconnect **server-side at the source**, making #2205's (large, hot-path) reducer-removal non-lossy. Disjoint from the broadcast/workflow/restore-cohort removals (they touch no server files) — can run fully in parallel.

armaxri:
## Summary

Phase-4 step-1 of #2139: begin making the shared `session-lifecycle` projection store **server-authoritative**, per the issue's corrected scope (the region is keyed by the **frontend tab id**, but backend sources know only the uuid `session_id`).

Delivered in two parts, plus a deliberate, reported deferral of the edges that cannot be made server-authoritative safely while the client dispatch is live.

### Part 1 — session_id → tab_id identity bridge (the real prerequisite)
- New `SessionManager.session_tab_ids` map (`session_id → tab_id`), populated in `create_connection` by parsing the tab id from `connect_id = ${tabId}:${retryCount}`, and cleared on **every** removal path (`close_session` and the natural-exit / dropped `emit_and_cleanup`).
- Public `tab_id_for(session_id)` lookup — the substrate the deferred drop/disconnect folds need.
- **All session kinds verified:** local / SSH / serial / telnet / docker / agent-remote (via `remote_proxy`) all flow through `create_connection` with `connect_id = ${tabId}:${retryCount}`. **Graphical / remote-desktop** sessions use a **separate** `GraphicalSessionManager` and a distinct `remote-desktop-state` lifecycle — they do **not** participate in the `session-lifecycle` region and dispatch no `session.*` intents, so they are correctly out of scope (not "unmappable"). The internal agent-setup session passes `connect_id = None` and is simply unmapped.

### Part 2 — server-side fold (safe slice only)
- New `fold_session_transition` (twin of `fold_monitor_transition`).
- Wired at the `create_connection` command for the **initial connect** (`retryCount == 0`): `connect` at entry, `connected` on success — keyed by the tab id read straight off `connect_id`. These converge exactly with the client's `session.connect` / `session.connected` dispatch, so both running is a benign convergent double-write; the coarse status field is not yet rendered from the region (that inversion is #2205).

### Deferred (would corrupt/diverge from the still-live client — reported per the de-risk instruction)
The client `session.*` dispatch stays on by default (`appStore` untouched, no reducer removed — that's #2205). The following source folds were investigated and **not** wired because they would double-write **conflicting** transitions against the live client:
- **`dropped` @ `terminal-exit`** — the backend `emit_exit` fires for every exit with no reason; the killed / clean / dropped classification is frontend-only (`consumeSessionKilled`, exit-code interpretation, `TerminalView` TCP-drop detection). Worse, a **resilient-reconnect** tab drives `session.reconnect` (→ Reconnecting), not `session.dropped` (→ Disconnected), and the backend cannot know a tab is resilient — a blind `dropped` fold would corrupt those.
- **`disconnect` @ `close_session`** — `close_session` cannot tell a **tab-close** (client dispatches `session.remove`) from a **kill** (client dispatches `session.disconnect`); folding `disconnect` races the `remove` and re-adds an orphan entry.
- **`connect_failed`** and **reconnect-attempt** folds — the frontend silently auto-retries an initial agent-connect failure (no `session.connectFailed` until MAX attempts), and reconnect attempts are owned by the client + backend timer (#2203).

These become safe once #2205 removes the client dispatch (or behind the mutation flag being off); a follow-up issue tracks them.

## Tests
- Identity-bridge unit tests: parse helper, record-on-create (with/without `connect_id`), cleanup on `close_session` and `emit_and_cleanup`.
- Server-authority projection-assertion tests (against `tauri::test::mock_app()`): backend-folded `connect` / `connected` create the **tab-id-keyed** region entry and fan one diff out with no client dispatch; parity test proves the fold reproduces the client `session.connect` route's snapshot exactly.

## Verification
`cargo build`, `cargo test --lib` (session + projection suites green), `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings` all clean. No frontend changes (CI `format:check` scope unaffected).

Closes #2431


armaxri:
Follow-up filed for the deferred server-side folds (dropped / disconnect / connect_failed): #2439.

armaxri:
Superseded by #2440 — recommitted with a commitlint-compliant (wrapped) body. Identical diff, no force-push.

Issue #2457, opened 2026-08-07
frontend re-attach of a backend-driven direct reconnect (new-session-id propagation)


armaxri:
Follow-up to #2454 (retention foundation PR). Discovered scoping the direct-connection backend reconnect redrive: completing #2454 needs a **frontend I/O re-attach** mechanism that does not exist today, and it is the real risk-bearing part of the redrive.

## Why
Today the reconnect **redrive** is client-owned: the #2203 backend timer fires only the `Waiting→Attempt` (Connecting) edge, then the frontend reconcile (`appStore.ensureSessionReconcileWired` → `reconnectTerminal` → `Terminal.tsx` → `createTerminal` → `create_connection`) re-establishes the transport, learns the **new backend `session_id`**, subscribes output/exit, and `setTabSessionId`s the tab.

To move the redrive server-side (the goal of #2454), the backend must create the connection itself. But on a genuine drop `SessionManager::emit_and_cleanup` tears down the `SessionEntry`, its output buffer, and the tab binding — so a direct reconnect necessarily mints a **new** backend `session_id`. With the client redrive suppressed (as #2454 requires — the reconnect engine is non-idempotent, so it must be a real authority cut), the frontend `Terminal` never learns that new id and terminal I/O breaks.

**So a backend redrive behind a flag is only safe once the frontend can re-attach to a backend-chosen session id.** Turning the flag on without this strands the ventilator reconnect hot path.

## Scope
- [ ] Propagate the current backend `session_id` per tab to the frontend (e.g. add `sessionId` to the shared `session-lifecycle` region record, keyed by tab id, or a dedicated signal).
- [ ] Give `Terminal.tsx` an **attach-to-given-session** path (generalize the existing `restartPersistentSessionForTab` / `initialSessionIdRef` reattach) that subscribes output/exit + `setTabSessionId` + replays local scrollback, **without** calling `create_connection`.
- [ ] Gate it behind the same new default-off flag as the backend redrive; when off, the client redrive path is unchanged.

## Done when
- With the flag ON, a real direct-SSH drop drives Reconnecting→Attempt→Connected fully server-side, terminal I/O re-attaches to the new backend session, and there is no double-count against the (suppressed) client mirror.
- With the flag OFF, behavior is identical to develop.

This is the ventilator reconnect hot path; verify locally against a real SSH drop (integration is not in per-PR CI). Sequenced with #2454's backend redrive wiring; the retention (Model A) foundation is already merged. See #2446 for the umbrella.

Issue #2455, opened 2026-08-07
backend reconnect redrive for agent silent-retry orchestration


armaxri:
Part 2 of #2446 (retention Model A). **Sequence after Part 1 (#2454)** — shares the default-off flag / client-mirror-suppression substrate; do not parallelise.

## Scope
- Move the **agent silent-retry** orchestration server-side (or model it as backend attempts): `MAX_AGENT_SPAWN_ATTEMPTS`, `connectRemoteAgent` transport re-establishment, and agent-transport tab parking — so one loop attempt ≠ one `create_connection` is handled at the source, not the client engine.
- Extend the Part 1 retained-request + redrive path to agent-hosted connections under the same flag, with the same drop+zeroize-on-terminal-state guarantee.

## Verification
- Local integration against a real agent drop/reconnect; confirm silent-retry attempts and parking behave as the client engine did, with the client mirror suppressed under the flag.

Once both parts land and are verified, #2446 closes and #2205 PR-B flips the flag + deletes the client engine. Reference: #2446 analysis.

armaxri:
**Implementation attempt (dev1) — concluded this needs sequencing, not a single flag-gated PR. No code landed; analysis below.**

Studied Part 1 (#2454/#2457/#2458, merged) and the agent reconnect subsystem to extend the same machinery to agent-hosted connections. The two reconnect paths do **not** compose as a drop-in, and the safe part is already merged:

**1. Agent tabs are architecturally outside Part 1's reconnect loop.** `isResilientReconnectTab` (`src/store/appStore.ts:2046`) excludes agent tabs (`if (cfg.agentId) return false`) and persistent tabs. Agent tabs therefore pass `resilient_reconnect=false` into `create_connection`, never enter the backend `session-lifecycle` reconnect loop / #2203 timer, and `AppReconnectRedrive::redrive` never fires for them. "Extend the same machinery" first requires **bringing agent tabs into that loop** — a flag-gated change to the core tab classification, broader than #2454's (which changed only the reconnect *drive*, not classification).

[...]

Issue #2446, opened 2026-08-07
Phase-4 step-4 (session-lifecycle): move the reconnect redrive loop server-side (unblocks #2205 PR-B)


armaxri:
Discovered completing #2439 (steps 1-3 merged/in-flight fold kill/disconnect/drop/reconnect/initial-connect-failed server-side). The final blocker for #2205 PR-B (deleting the appStore reconnect engine) is that the **reconnect-loop give-up is not source-foldable** while the loop orchestration is client-owned. Per the step-3 analysis (#2445):

The #2203 backend timer drives only `Waiting→Attempt`; each attempt's **redrive** (`reconnectTerminal`→`Terminal.tsx`→`createTerminal`), the **agent silent-retry** loop (`MAX_AGENT_SPAWN_ATTEMPTS` + agent-transport parking, so one loop attempt ≠ one `create_connection`), and **cancellation semantics** all live in the client engine (`driveAutoReconnect` `appStore.ts:2199-2331` + `reconnectReducer`). Until these move server-side, deleting the client engine would strand the backend loop at `Connecting`.

## Scope (the agent-centric core — sequence carefully, integration-verify)
- [ ] Move the per-attempt connection **redrive** server-side (backend re-establishes the transport on the timer's `Attempt` edge, folds `connected`/`reconnectFailed` per attempt).
- [ ] Move the **agent silent-retry** orchestration server-side (or model it as backend attempts).
- [ ] Move **cancellation** semantics server-side (user-cancel vs give-up).
- [ ] Fold the final **give-up → disconnected** at the source.
- [ ] Then #2205 PR-B is a mechanical removal of the now-redundant client engine + appStore slice.

This is the largest/most safety-critical piece of the session-lifecycle inversion (ventilator reconnect hot path) and needs release-cadence/local integration verification. **Blocks #2205 PR-B.**

armaxri:
Labelled `Ready2Implement` per maintainer (2026-08-07): queued as the next single-slot priority — this is the critical-path unblocker for #2205 PR-B → #2206 → closes #2152. Steps 1-3 (#2443/#2444/#2445) are merged, so the prerequisites are in place. Will be worked on its own slot (not paired) given it's the ventilator reconnect hot path and needs local/release-cadence integration verification.

armaxri:
**Blocked on a maintainer architecture/security decision — removing `Ready2Implement` (not implementable as-is), per the #1351 discipline.** Analysis from a scoping pass:

[...]

Issue #2437, opened 2026-08-07
flaky test: agent persistent_shell_buffer_replayed_after_tcp_reconnect (re-attach 'Session not running')


armaxri:
Recurring **agent integration** flake, unrelated to the PR under test (hit on #2436, a frontend-only restore-cohort change that touches nothing in `agent/`).

```
thread 'persistent_shell_buffer_replayed_after_tcp_reconnect' panicked at agent/tests/local_agent_integration.rs:1287:9:
re-attach failed: {"error":{"code":-32001,"data":{"session_id":"..."},"message":"Session not running"}}
test result: FAILED. 11 passed; 1 failed
```

The re-attach races the daemon-backed session's readiness — the client re-attaches before (or after) the session is registered/still-running, getting `-32001 Session not running`. Timing-sensitive; blocks unrelated PRs on the ubuntu Run Tests leg.

## Scope
- [ ] Make the re-attach deterministic: wait for the session to be registered/running (poll `is_alive`/list with a bounded timeout) before asserting the buffer replay, rather than re-attaching on a fixed delay.
- [ ] Confirm no real re-attach regression (the buffer *is* eventually replayed; only the timing races).

Part of the recurring-flake sweep (cf. #2395, #2194).

PR #2458, opened 2026-08-07
feat(session): retain resilient direct connection requests for reconnect redrive (part of #2454)

### LINKED ISSUE #2454: backend reconnect redrive (direct connections) behind a default-off flag with client-mirror suppression
Part 1 of #2446 (retention Model A chosen — retain the full connection request in-memory per tab). Unblocks #2205 PR-B.

## Scope
- Retain the connection request (`type_id`+`settings`+`agent_id`+`resilient`) in-memory per tab when a resilient session is established, for **direct (non-agent) connections**. Drop+zeroize it on any terminal loop state (give-up / user-cancel / success), so resolved secrets never outlive the session.
- On the #2203 backend timer's `Waiting→Attempt`(Connecting) edge, have the **backend** re-establish the transport from the retained request and fold `connected`/`reconnectFailed` per attempt at the source (today `create_connection` folds outcomes only for `retryCount==0`).
- Put this behind a **new default-off flag** that also **suppresses the client redrive + client mirror** (`ensureSessionReconcileWired`/`session.reconnectFailed`) when on — the reconnect engine is non-idempotent, so this must be a real authority cut, not an additive convergent double-write.
- User-cancel-vs-give-up and give-up→disconnected are already modeled in the store and become source-owned for free once redrive is source-owned; verify they fold correctly.

## Verification (ventilator hot path — integration NOT in per-PR CI)
- Local: `./scripts/dev.sh` + the `tests/system` SSH-disconnect suite; confirm with the flag ON a real drop drives Reconnecting→Attempt→Connected, and give-up→Disconnected, with no double-count against the (suppressed) client mirror.
- Unit + client-route parity tests as #2443/#2444/#2445 did.

Agent silent-retry orchestration is deliberately excluded — see Part 2. Reference: concept `docs/concepts/future/stateless-ui-agent-centric.html` Phase 4; #2446 analysis.

### LINKED ISSUE #2459: flaky test: daemon detach_keeps_session_alive_when_daemon_closes_connection times out under macOS CI load
The regression test added by the #2437 fix (PR #2450), `daemon::client::tests::detach_keeps_session_alive_when_daemon_closes_connection` in `agent/src/daemon/client.rs`, is itself timing-fragile and **false-failed on macOS CI on an unrelated PR** (#2458, which touches only `src-tauri/src/session/*` — no `agent/` code). The ubuntu leg passed the identical suite on the same commit, and a re-run cleared it — so it's a macOS-under-load timing flake, not a real regression.

## Root cause
The test does a fixed `tokio::time::sleep(Duration::from_millis(100))` after `detach()` and then asserts `is_alive()` / `!exit_ran` (`client.rs` ~775-785). Under a loaded macOS runner the mock-daemon reader/abort interplay can slip past that fixed window, so the assertion races the very ordering it's trying to prove. It's the same fixed-delay anti-pattern the #2437 *integration* flake had — the fix hardened the source but left the new unit test on a fixed sleep.

## Scope
- Replace the fixed 100ms sleep with a deterministic wait: drive the mock daemon to a known point and assert on an observable signal (e.g. poll that the reader task has been aborted / that no exit-hook fired within a bounded, generous timeout) rather than a single fixed delay.
- Confirm it still fails on the pre-#2437 ordering (the regression guard must stay real) and passes deterministically with the fix, including a stress loop.

Part of the recurring-flake sweep (cf. #2437, #2395, #2194). Low urgency but it blocks unrelated PRs' macOS leg.

armaxri:
Part of #2454 (retention **Model A**, part of #2446; unblocks #2205 PR-B). Lands the **safe, inert foundation** for the direct-connection backend reconnect redrive. **Does not close #2454** — the redrive itself and its frontend re-attach are deferred (see *Deferred* below), because completing them cleanly requires a new frontend I/O re-attach mechanism that does not exist yet (#2457) and can only be verified by local/release-cadence integration on the ventilator reconnect hot path, not per-PR CI.

## What moved server-side
- New `src-tauri/src/session/retained_request.rs`: `RetainedConnectionRequest` (`type_id` + `settings` + `agent_id` + `resilient`) and a per-tab `RetainedRequestStore`, keyed by the frontend **tab id** (the stable key that survives a reconnect, matching the `session-lifecycle` region and the `session_tab_ids` identity bridge).
- `SessionManager::create_connection` now **retains** the request for a resilient **DIRECT** (non-agent) session — one uniform Model A path for saved / ad-hoc / spawned / inline connections. Agent silent-retry (#2455) and non-resilient tabs never retain.
- The request is dropped + **zeroized** on every path where the tab stops being active/reconnecting:
  - `SessionManager::close_session` (user close of a live session),
  - and the lifecycle-intent routes in `session_projection::projection` — `session.disconnect` (graceful), `session.cancelReconnect` (user-cancel), `session.reconnectFailed` **when it exhausts the loop** (give-up → Failed), and `session.remove` (tab/session gone).

## The flag / default-off
No behavior flag is introduced in this PR, because it lands only the retention foundation, which is **inert**: nothing reads the retained request back yet, and it is only ever populated for a resilient direct session. Develop's behavior is therefore unchanged. The new default-off flag (backend redrive + client-mirror/redrive suppression) belongs with the redrive wiring — see *Deferred*.

## Zeroize-on-terminal-state (mandatory secret mitigation)
`settings` can carry resolved secrets (SSH password/passphrase after credential-store + jump-host resolution). Model A extends that secret's in-memory lifetime past the drop for the loop's duration, so `RetainedConnectionRequest::drop` recursively **zeroizes** every string leaf of `settings` (plus `agent_id`/`type_id`) via the same `zeroize` crate the credential store uses. Every terminal-loop / session-end path above drops + zeroizes, so no resolved secret outlives an active or reconnecting session.

## Deferred (why this doesn't complete #2454)
On a genuine drop, `emit_and_cleanup` tears down the `SessionEntry`, its output buffer, and the tab binding — so a backend-owned direct reconnect necessarily mints a **new** backend `session_id`. With the client redrive suppressed (required — the reconnect engine is non-idempotent, so #2454 must be a real authority cut, not an additive double-write), the frontend `Terminal` has no way to learn that new id and re-attach terminal I/O. That frontend re-attach / new-session-id propagation is a distinct, integration-verified prerequisite, filed as **#2457**. Once it lands, the remaining #2454 work is: wire the #2203 timer's `Attempt` edge to redrive from the retained request behind the new default-off flag + fold `connected`/`reconnectFailed` at the source + suppress the client redrive/mirror when on.

## Verification
- `cargo test --lib` (retention store, zeroize, and manager retain/clear paths): green — new tests in `retained_request.rs` and `session::manager::tests` (`create_connection_retains_request_for_resilient_direct_session`, `..._does_not_retain_for_non_resilient_session`, `..._without_tab_retains_nothing`, `close_session_clears_the_retained_request`, `clear_retained_request_is_idempotent`, plus `zeroize_json_scrubs_nested_string_leaves` / `drop_zeroizes_secret_settings`).
- `cargo clippy --lib --all-features -- -D warnings`: clean. `cargo fmt --check`: clean.
- No integration run needed: this change is behaviorally inert (no consumer reads the retained request), so there is nothing on the reconnect hot path to exercise yet. The redrive that does touch the hot path is deferred to #2457 + the #2454 follow-up, where local SSH-drop verification is mandatory.

Follow-up: #2457.

---

Closes #2459 — hardened the flaky `detach_keeps_session_alive_when_daemon_closes_connection` reader-abort race test (was failing the macOS `Run Tests` leg on a pre-existing flake unrelated to this PR). The fixed 100ms sleep is replaced with deterministic synchronization: the mock daemon now closes the connection only after the client reader task is provably gone, matching production ordering. Test-only change; no production code touched. Guard verified still to fail on the pre-#2437 ordering.

### DIFF
--- a/agent/src/daemon/client.rs
+++ b/agent/src/daemon/client.rs
@@ -741,12 +741,27 @@ mod tests {
         );
         let endpoint = transport::session_endpoint(&session_id);
 
-        // Mock daemon: accept, send Ready, then close the connection the instant
-        // it receives MSG_DETACH — exactly what the real daemon does, EOFing the
-        // client's reader.
         let mut listener = transport::DaemonListener::bind(&endpoint)
             .await
             .expect("bind mock daemon");
+
+        // Mock daemon. The real daemon only closes the connection *in response* to
+        // MSG_DETACH, so the resulting EOF reaches the client well after detach's
+        // (asynchronous) reader-task abort has landed. The original mock instead
+        // dropped its halves the instant it saw MSG_DETACH — a tighter schedule
+        // than production that raced the abort and made this test flaky on macOS
+        // CI (#2459). Reproduce production's ordering *deterministically*: do not
+        // EOF the client until its reader task is actually gone. The reader task
+        // owns the client's read half, so once a correct detach aborts it (and
+        // detach has dropped the client's write half) the client socket is fully
+        // closed and any write from us fails; probe for exactly that. While the
+        // reader still lives, empty MSG_OUTPUT frames are accepted and ignored.
+        //
+        // `reader_gone` reports whether the reader was gone *before* we closed —
+        // the direct #2437 property. A correct detach makes it true quickly; the
+        // pre-#2437 ordering leaves the reader live, so the probe never fails and
+        // the generous deadline trips it to false.
+        let (reader_gone_tx, reader_gone_rx) = tokio::sync::oneshot::channel::<bool>();
         let server = tokio::spawn(async move {
             let (mut reader, mut writer) = listener.accept().await.expect("accept");
             protocol::write_frame_async(&mut writer, MSG_READY, &[])
@@ -757,7 +772,23 @@ mod tests {
                     break;
                 }
             }
-            // Drop both halves → the client's reader half observes EOF.
+            let deadline = tokio::time::Instant::now() + Duration::from_secs(2);
+            let reader_gone = loop {
+                if protocol::write_frame_async(&mut writer, MSG_OUTPUT, &[])
+                    .await
+                    .is_err()
+                {
+                    break true;
+                }
+                if tokio::time::Instant::now() >= deadline {
+                    break false;
+                }
+                tokio::time::sleep(Duration::from_millis(1)).await;
+            };
+            let _ = reader_gone_tx.send(reader_gone);
+            // Only now drop our halves. On a correct detach the client's reader is
+            // already gone, so this EOF reaches no one; on the pre-#2437 ordering
+            // it is what finally trips the still-live reader's exit path.
             drop(writer);
             drop(reader);
             listener.cleanup();
@@ -773,9 +804,16 @@ mod tests {
 
         client.detach().await;
 
-        // Give any stray reader activity a chance to (wrongly) fire first.
-        tokio::time::sleep(Duration::from_millis(100)).await;
-
+        // Deterministic synchronization (replaces the old fixed 100 ms sleep): the
+        // mock only closes the connection once the reader task is gone, so it can
+        // never deliver a detach-induced EOF to a live reader.
+        let reader_gone = reader_gone_rx
+            .await
+            .expect("mock daemon reported reader state");
+        assert!(
+            reader_gone,
+            "detach must abort the reader before the daemon connection closes (#2437)"
+        );
         assert!(
             client.is_alive(),
             "a clean detach must not mark the still-alive session dead (#2437)"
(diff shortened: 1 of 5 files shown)


armaxri:
CI note: the macOS `Run Tests` leg is red on an **unrelated** pre-existing flake — `daemon::client::tests::detach_keeps_session_alive_when_daemon_closes_connection` (agent/), which this PR does not touch (diff is `src-tauri/src/session/*` only). Ubuntu passed the identical suite; it fails deterministically only on macOS (non-synchronous `tokio` abort racing the detach-EOF under load). Folding the #2459 hardening for that test into this PR so its macOS leg can go green; this PR will then also `Closes #2459`.

Issue #2480, opened 2026-08-08
live agent-reconnect harness: agent registry-daemon handshake stalls over a throwaway sshd (blocks #2476 live grade)


armaxri:
Follow-up to #2476 (PR forthcoming). This is the remaining blocker to **live-verifying** the agent reconnect activation.

## Context
#2476 adds `LocalAgentSshd` (a harness-controlled throwaway loopback sshd with the `termihub-agent` binary reachable — the test-owned equivalent of `scripts/dev.sh`'s dev agent) and `tests/system/tests/test_agent_reconnect_live.py`, which drives a real agent-transport drop (kill the sshd) and reconnect.

The scenario gets **almost all the way** against the built app:
1. app launches with the seeded key-auth agent + experimental UI,
2. connect is driven by the agent's known id,
3. SSH connects, **key auth succeeds**, **host-key pre-trust works** (seed `~/.ssh/known_hosts`),
4. the **agent binary starts over SSH** (`termihub-agent 0.1.0 starting in stdio mode` → `Stdio transport loop started` → `Spawned registry daemon`).

Then it stalls: the agent connection never reaches `connected`. The handshake appears to hang after `Spawned registry daemon` (ADR-11 host-wide registry daemon) over the minimal throwaway sshd, and concurrently the WKWebView is occlusion-throttled (#957/#2460) while the harness terminal is focused, so bridge commands time out (60 s).

## Scope
- [ ] Determine why the agent registry-daemon handshake does not complete over the throwaway sshd (vs. the deployed-agent container) — e.g. the daemon spawn/bind step, or a stdio-channel assumption; consider an agent option to skip/inline the registry daemon for a single-client stdio session in tests.
- [ ] Mitigate webview throttling for the long agent connect (extend the anti-occlusion handling / raise the live request timeout for the agent lane).
- [ ] Un-gate `test_agent_reconnect_live.py` (currently skips unless `TERMIHUB_TEST_LOCAL_AGENT_RECONNECT=1`) and green **both** lanes: flag-off (client-driven) and flag-on (backend-driven, prolonged drop → reattach; give-up → disconnected).
- [ ] Depends on #2479 (projection-based agent/settings harness reads).

## Done when
`test_agent_reconnect_live.py` runs green unattended in both lanes, completing the #2476 live verification (backend-driven reconnect across a prolonged drop, client engine suppressed, no double-drive, no stranded tab, give-up → disconnected).

armaxri:
[...]

Issue #2476, opened 2026-08-08
activate agent reconnect redrive: flag-gate the frontend routing + suppress the client agent engine


armaxri:
Follow-up to #2473 (Part 2 of #2446). The **backend foundation** for the agent activation landed in PR for #2473 (retention for resilient agent sessions + the per-agent transport-config refcount scrub — both inert with the flag off). This issue is the remaining **risky, all-or-nothing frontend activation**, deliberately split out because it must land as one coherent cut *and* be verified live against a real agent drop before it can be trusted (the ventilator hot path).

## Why this was split from #2473

The backend half is safe to land alone (fully inert: no agent session is resilient and no agent config is retained until this frontend cut routes them). The frontend cut is the opposite — it *activates* the agent reconnect hot path under the flag, and a partial/unverified cut makes flag-ON **broken** for agents. Per the ventilator-grade guardrail it was not landed without live verification.

## Scope (land as ONE coherent cut)
- [ ] **Flag-gate `isResilientReconnectTab`** (`src/store/appStore.ts` ~2046) to include agent tabs only when `sessionBackendReattach` is on. Flag-off keeps agents excluded → byte-identical.
- [ ] **Thread `resilientReconnect` + `backendReattach` through the `remote-session` branch of `createTerminal`** (`src/services/api.ts` ~207) — it currently DROPS both args for agent sessions (`createConnection(sessionType, rest, agentId, connectId)`), so the backend never sees them for agents. This is required for the backend retention to fire.
- [ ] **Suppress the client agent silent-retry** for backend-driven reconnects (`src/components/Terminal/Terminal.tsx` ~631-711): under the flag, on a reconnect (retryCount>0) the backend is the sole driver, so the client must NOT run the `connectRemoteAgent` + park + bounded `MAX_AGENT_SPAWN_ATTEMPTS` engine (double-drives the transport the redrive is re-establishing). The outcome-mirror suppression in `driveAutoReconnect` already covers agents once they are resilient (it gates on the flag, not the tab type).
[...]

Issue #957, opened 2026-06-30
Operator run-through of the external-app guided-manual suite (test_external_app.py)


armaxri:
## Goal

The guided-manual suite `tests/system/tests/test_external_app.py` was **authored and statically verified** under #917 (collects, skips cleanly on CI, every testid/mixin cross-checked against source). It has **not** yet been run through the full operator loop, because that needs a GUI machine with the app built, Docker SSH fixtures up, an interactive operator, and — for X11/clipboard — a local X server.

This ticket tracks that end-to-end run so #917 can close on "tests authored" without the verification being forgotten.

## To do

Run on a GUI host (per machine roles, e.g. macOS with XQuartz, and a Linux box):

```bash
# from tests/system/, with Docker SSH fixtures available
./pytest.sh --manual -k external_app -s
```

Verify each item and capture the generated `manual-<ts>-<platform>-<arch>.{json,md}` report:

- **Open in VS Code** — local (MT-FB-04/14), SFTP (MT-FB-15, no crash #828), menu hidden when VS Code absent (MT-FB-16).
- **SSH agent auth** — connect with `agent` auth (MT-SSH-07); Setup-Agent launcher (MT-SSH-09, currently via the error dialog — see #955).
- **X11 forwarding** — flag persists, session connects, window appears (MT-SSH-14/15/16/18, MT-XPLAT-03). Needs a local X server.
- **Clipboard** — copy/paste via tab menu + platform shortcuts (MT-KB-01..04).

## Possible enhancement found while authoring

`test_x11_forwarding_window_appears` leaves the `$DISPLAY` check to the operator because the SSH fixture container's X11 support is unverified. If a fixture image with `X11Forwarding yes` + `xauth` is available, the test could **auto-assert** that `$DISPLAY` is set (e.g. `localhost:NN.0`) and only leave the visible window to the operator.

Surfaced during #917. Part of epic #913.

armaxri:
## Closing — operator run complete; X11 forwarding verified working

All code blockers surfaced by this operator run are fixed and merged to `develop`:

- **#1304** (PR #1308) — X11 forwarding allocated a huge ephemeral display number (`:26961`), so stricter X clients refused it and the forwarded channel never reached the forwarder. Now allocates a conventional display (`:10`, OpenSSH-style).
[...]

PR #2489, opened 2026-08-08
add headless fresh-after-reconnect create coverage; agent registry handshake does not stall (part of #2476, #2480)

armaxri:
## Summary

Headless automated coverage for the ventilator agent-reconnect hot path behind the maintainer's live failure (#2476: *"agent transport restored, but the tab stayed stuck Reconnecting, no connection came back"*), and the #2480 hypothesis that the ADR-11 registry-daemon handshake stalls for a fresh-after-reconnect agent.

The agent crate runs a **real agent binary over a real local transport with no GUI/webview**, so — unlike the display-blocked manual grade — this reproduces and auto-tests the scenario headlessly.

## Did it reproduce in the agent suite? **No — it does not stall.**

New test `fresh_agent_after_reconnect_creates_session_over_surviving_registry` (`agent/tests/local_agent_integration.rs`):

1. Agent A connects, creates a live shell session (this spawns the detached registry daemon), and the test **confirms the registry endpoint is genuinely reachable**.
2. Agent A is dropped (process killed) — the headless analog of the transport drop taking the remote agent with it. The detached registry daemon survives (asserted).
3. Agent **B** — a *fresh* process, the analog of the reconnect standing a new `termihub-agent` up over the re-established transport — is pointed at the **same, now pre-existing** registry endpoint. It must complete `initialize` + a fresh `connection.create` and reach a usable session.

**Result:** B's `initialize` + `connection.create` completes and the shell echoes in **~2 s** over a genuinely surviving registry daemon. Stable across 7 consecutive runs. A real stall would FAIL the test (the 30 s RPC read timeout, or the bounded create-elapsed ceiling) rather than hang.

## Root cause / finding

The agent's registry client is **non-blocking and infallible by contract**: `initialize` registers the client fire-and-forget (`agent/src/handler/dispatch.rs`), and the supervisor (`agent/src/registry_daemon/client.rs`) owns all connect/spawn/retry in the background — so the registry-daemon handshake is **never on the `connection.create` path**. The detached-spawn path (`agent/src/daemon/spawn.rs`) also null's the daemon's stdin/stdout and redirects stderr to a log file, so there is no SSH-channel fd leak by construction.

This test proves that end-to-end over a real transport. It therefore **narrows the live "stuck Reconnecting" out of the agent side** (registry logic + fresh create) and toward either the desktop↔agent transport-reconnect layer (`src-tauri/src/terminal/agent_manager.rs` `reconnect_agent` settling / it triggering the redrive's create) or the webview occlusion-throttle that #2480 itself names as the second cause. The #2480 stall observed over the throwaway sshd is specific to that stdio-over-SSH transport path, not the agent's registry handshake.

## Also: harness isolation fix

The integration harness set `XDG_CONFIG_HOME` but **not** `TERMIHUB_REGISTRY_ENDPOINT`, so its agents spawned/joined the **shared per-user** registry (`/tmp/termihub/{user}/registry.sock`) — leaking a detached process into shared state and letting parallel checkouts fight over one registry. Each agent now gets a per-agent endpoint (mirroring `registry_daemon_integration.rs`); the surviving-registry test relies on this to own a shared endpoint across the agent swap.

## What still needs a display-backed run

- The **desktop-side** transport-reconnect settling (`agent_manager::reconnect_agent`) + it driving the redrive's fresh create over a **real re-established SSH transport** — that requires a real sshd (the #2480 `LocalAgentSshd` system-harness lane), not a headless unit.
- The **webview occlusion-throttle** mitigation (#957/#2460/#2480) is inherently GUI and out of scope here.

## Test plan

- `cargo test -p termihub-agent --test local_agent_integration` — 13 passed (new test + all pre-existing), stable across repeated runs.
- `cargo fmt -p termihub-agent -- --check` and `cargo clippy -p termihub-agent --tests -- -D warnings` clean.

Part of #2476, #2480.

### DIFF
--- a/agent/tests/local_agent_integration.rs
+++ b/agent/tests/local_agent_integration.rs
@@ -84,7 +84,33 @@ impl LocalAgent {
         // live in which a connection could be reset.
         let config_dir = TempDir::new().expect("failed to create temp config dir");
 
-        let (process, addr, stderr) = spawn_ready_listener(config_dir.path(), None);
+        // Point the agent's host-wide registry (ADR-11) at a per-agent endpoint
+        // inside this temp dir rather than the shared per-user default
+        // (`/tmp/termihub/{user}/registry.sock`). Without this override the agent
+        // spawns/joins the developer's real registry daemon — leaking a detached
+        // process into shared state and letting parallel checkouts fight over one
+        // registry. See [`registry_endpoint_in`].
+        let registry_endpoint = registry_endpoint_in(config_dir.path(), "reg");
+        let (process, addr, stderr) =
+            spawn_ready_listener(config_dir.path(), &registry_endpoint, None);
+        LocalAgent {
+            process,
+            addr,
+            _config_dir: config_dir,
+            _stderr: stderr,
+        }
+    }
+
+    /// Spawn an agent pointed at an explicit, caller-owned registry endpoint.
+    ///
+    /// Used by the fresh-after-reconnect test to stand a second agent process up
+    /// against a registry daemon a **prior** agent already spawned — the headless
+    /// analog of a reconnect re-establishing the transport with a fresh
+    /// `termihub-agent --stdio` while the host-wide registry survives the swap.
+    fn spawn_with_registry(registry_endpoint: &str) -> Self {
+        let config_dir = TempDir::new().expect("failed to create temp config dir");
+        let (process, addr, stderr) =
+            spawn_ready_listener(config_dir.path(), registry_endpoint, None);
         LocalAgent {
             process,
             addr,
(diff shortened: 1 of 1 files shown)


armaxri:
Windows leg is blocked by the #2490 connect-timeout flake (two runs, two *different* agent-integration tests, same `os error 10060` — ubuntu+macos green). Holding this PR until #2490's deterministic connect-retry lands, then rebasing + re-running.

armaxri:
Held: this PR (agent-side 'does-not-stall' coverage + registry isolation) keeps failing the **Windows** leg on the chronic agent-integration flake (#2495) — a THIRD facet after the connect (#2492) and read (#2494) fixes: the agent runs >60s without responding on the Windows runner, then 10060s. ubuntu+macOS green. **Not blocking anything critical** — the real reconnect fix (#2491) and its headless coverage are already merged; this PR is supplementary coverage. Holding it pending the #2495 Windows-lane decision (quarantine-on-Windows vs deeper agent-startup fix) rather than burning more reruns.

Issue #2460, opened 2026-08-07
live-SSH-connect integration suite cannot be driven to green via the bridge harness (blocks #2398 live half)


armaxri:
Follow-up to #2398. While verifying the transfer-queue integration suite against the region-authoritative panel (#2229), the **region half passed reliably** (`TestTransferQueuePanel`, 7/7, across multiple runs) but the **live SFTP sub-suite** (`TestTransferQueueLiveTransfer`) could not be driven to green on the macOS bridge harness — and the blocker is **not** in the transfer-queue code. It reproduces identically on `test_ssh.py::TestSshPasswordAuth`, so it is systemic to the **live-SSH-connect flow driven over the bridge**, independent of #2229/#2398.

### Symptom
Across 5 runs (`test_transfer_queue.py` live suite ×3, `test_ssh.py::TestSshPasswordAuth` ×1, a diagnostic ×1) the live SSH connect is **never initiated** — the app log always stops after `Loading connections and folders` with **no SSH/session backend activity at all**. The failure point *varies*:

- Sometimes at `create_ssh_connection` → `select_connection_type("ssh")`: the bridge `select` verb returns `command timed out after 10.0s` (the app JS thread does not ack within 10s).
- Sometimes later at `handle_password_prompt`: `timed out waiting for the SSH password prompt` — the editor is fully filled (verified via the failure-bundle screenshot: TYPE=SSH, HOST=127.0.0.1, PORT=2201) but Save & Connect never opens a session.

The recurring **10s bridge-command timeouts** (the app not acking a command for 10+ seconds) point at **webview unresponsiveness / host contention** on this machine, not a deterministic code path.

### Already addressed (PR #2461)
The "Save & Connect never opens a session even though the form is filled" case has a concrete root cause: the editor buttons are `aria-disabled`/`data-invalid` (not natively `disabled`), so a bridge click still dispatches but `handleSaveAndConnect` early-returns on `!canSave`; when the harness clicks before async form validation settles, the click is a silent no-op. PR #2461 makes `create_ssh_connection` wait for the form to report itself valid (`data-invalid` cleared) before clicking. **This fix could not be validated end-to-end** because the 10s `select_connection_type` jank now fails the run *earlier*, before the save-connect step is reached.

### Scope
- [ ] Reproduce the live-connect integration failure on a **quiet** machine (no concurrent cargo builds / other test-app instances) to separate host-contention from a real regression.
[...]

Issue #2490, opened 2026-08-08
flaky: Windows agent-integration TCP connects time out (os error 10060) across multiple tests


armaxri:
Recurring **Windows-only** flake cluster in the agent integration suite: agent-integration tests intermittently fail with `Os { code: 10060, kind: TimedOut }` ("connection attempt failed because the connected party did not properly respond") on the `Run Tests (windows-latest)` leg, while ubuntu + macos pass the identical test. It reruns green, so it's timing/flake, not a logic defect — but it blocks unrelated PRs' Windows leg and costs a rerun each time.

## Observed instances
- `shell_session_persists_across_client_disconnect` (`agent/tests/local_agent_integration.rs:~680`, `read_line` connect) — blocked PR #2489.
- `listening_log_is_a_true_readiness_signal` (`agent/tests/tcp_listener_readiness.rs`) — blocked PR #2470; tracked at #1579.
- Same 10060 signature seen on other agent-integration reruns this session.
Related earlier flake hardening: #2437, #2459 (detach race — deterministic-wait fixes).

## Likely cause
The tests connect to a freshly-bound loopback TCP listener (agent `--listen`) and read the first line, but on Windows CI the connect races the listener actually accepting — a fixed/implicit timeout window that's too tight under Windows CI load, so the connect times out at 10060 instead of retrying.

## Scope
- Make the agent-integration TCP client connect **retry with a bounded backoff** until the listener accepts (or a generous deadline), rather than a single connect that 10060s under load — deterministic wait on readiness, not a fixed window (mirror the #2459 approach).
- Apply to the shared connect/readiness helper so ALL agent-integration tests benefit (`local_agent_integration.rs`, `tcp_listener_readiness.rs`), and fold in / close #1579.
- Confirm no real regression: the readiness signal must still be genuine (don't just swallow a never-accepting listener).

## Done when
The agent-integration suite's Windows leg stops intermittently 10060-ing on connect; a stress loop of the affected tests on Windows is green; #1579 is resolved by the shared fix.

PR #2491, opened 2026-08-08
fix(agent): fast-fail dead-but-lingering daemon socket so reconnect settles (part of #2476, #2480)

armaxri:
## Summary

Chases the **last un-covered layer** of the live agent-reconnect bug ("transport restored but the tab stayed stuck Reconnecting", #2476/#2480): the desktop backend re-establishing the **real russh transport** after the sshd returns, and whether it then drives a fresh session create. No test exercised real russh reconnect at this layer.

**Reproduced at this layer? YES — and it found a real, fixable bug (not the webview throttle).**

## Root cause

A new headless integration test (`src-tauri/src/terminal/agent_manager.rs`) stands up a **real local sshd** with the real `termihub-agent` reachable (the Rust analog of the Python `LocalAgentSshd`, #2481), drives `reconnect_agent`, kills the sshd for a genuine server-side transport drop, restores it, and asserts the reconnect re-establishes the transport and a fresh `connection.create` yields a usable session.

The test proved the desktop **`reconnect_agent` russh layer is correct** — it re-establishes the transport and drives a fresh create fine. But it surfaced a stall one layer deeper:

- Killing the sshd process tree also kills the **setsid'd session daemon** (#995), leaving its unix **socket file behind**.
- On reconnect, the fresh agent's startup `recover_sessions` re-attaches persisted sessions. Its liveness gate `endpoint_alive()` only checks the socket **file exists**, so a dead-but-lingering socket passed the gate.
- It then connected via the **30 s spawn-path** connect timeout, which retries `ConnectionRefused` for the full 30 s — and this runs **before** the stdio loop answers `initialize`.
- So the desktop's `reconnect_agent` sat blocked **~31 s per dead session** waiting for `initialize` — the exact "transport restored but stuck Reconnecting" symptom.

The spawn path can afford 30 s because it **races the connect against the daemon process exiting**; recovery has no process to race. Notably, the maintainer's own psutil recursive-kill harness (#2481) destroys the setsid'd daemon the same way this test does, so the live harness would hit this too.

## Fix

Give recovery a short connect timeout so a dead-but-lingering socket fast-fails and the session is dropped, instead of stalling startup:

- `agent/src/daemon/transport.rs`: `RECOVERY_CONNECT_TIMEOUT` (2 s) + `connect_for_recovery`.
- `agent/src/daemon/client.rs`: `DaemonClient::connect_for_recovery` (spawn + attach keep the 30 s timeout, unchanged).
- `agent/src/session/manager.rs`: `recover_sessions` uses the recovery variant.

## RED → GREEN

- **Integration test** (`reconnect_agent_reestablishes_russh_transport_and_drives_fresh_create`, real sshd + agent, headless): pre-fix `reconnect_agent` took **~31 s** (RED against the new 20 s `RECONNECT_SETTLE_CEILING`); post-fix it settles in **~4 s** (GREEN). Skips gracefully when `sshd`/`termihub-agent` are absent.
- **Hermetic regression** (`connect_for_recovery_fast_fails_dead_lingering_socket`, `agent/src/daemon/transport.rs`): `connect_for_recovery` against a dead-but-lingering socket fails in ~2 s, well within the 30 s spawn-path timeout. Runs in the normal agent test suite.

Both deterministic across repeated runs; poll-on-state, no fixed sleeps. `cargo fmt --check` + `cargo clippy --tests -- -D warnings` clean on both touched crates; agent daemon (61) + session (120) tests green.

## Test isolation

Headless only — no GUI/`dev.sh`/`tauri dev`, no Docker. The test agent is pointed at a per-test `TERMIHUB_REGISTRY_ENDPOINT` + `XDG_CONFIG_HOME` inside its own temp dir via sshd `SetEnv` (per-agent registry isolation, #2489); it kills only its own spawned sshd subtree (no pattern-pkill).

## Per-issue status

- **#2476** — a real reconnect stall on this layer is now fixed with automated coverage. The umbrella's frontend activation cut + live-grade verification are **not** in scope here, so this is **part of #2476**, not a close.
- **#2480** — this is a *different* stall from the registry-daemon handshake (#2489 handled that); it removes another blocker to the live grade. **Part of #2480**; the registry-handshake item and un-gating `test_agent_reconnect_live.py` remain owned there.
- **#2460 / #957** (webview occlusion throttle) — **not** implicated by this finding: the reconnect stall was real backend/agent behavior, reproduced headlessly with no webview involved. The throttle remains its own concern, owned by #2460/#957.

part of #2476, #2480

### DIFF
--- a/agent/src/daemon/client.rs
+++ b/agent/src/daemon/client.rs
@@ -99,6 +99,35 @@ impl DaemonClient {
         endpoint: String,
         notification_tx: NotificationSender,
     ) -> Result<Self, anyhow::Error> {
+        Self::connect_inner(session_id, endpoint, notification_tx, false).await
+    }
+
+    /// Connect to an **already-running** daemon during session recovery, failing
+    /// fast on a dead daemon whose socket file merely lingers.
+    ///
+    /// Like [`connect`](Self::connect) but uses the short recovery connect
+    /// timeout ([`transport::connect_for_recovery`]) instead of the long
+    /// spawn-path one. The spawn path races its 30s connect against the daemon
+    /// process exiting; recovery has no such process to race, so retrying a
+    /// dead-but-lingering socket for 30s would stall the fresh agent's startup —
+    /// and therefore the desktop's `initialize` handshake — after a reconnect
+    /// (#2476). This is the entry point [`recover_sessions`] must use.
+    ///
+    /// [`recover_sessions`]: crate::session::manager::SessionManager::recover_sessions
+    pub async fn connect_for_recovery(
+        session_id: String,
+        endpoint: String,
+        notification_tx: NotificationSender,
+    ) -> Result<Self, anyhow::Error> {
+        Self::connect_inner(session_id, endpoint, notification_tx, true).await
+    }
+
+    async fn connect_inner(
+        session_id: String,
+        endpoint: String,
+        notification_tx: NotificationSender,
+        for_recovery: bool,
+    ) -> Result<Self, anyhow::Error> {
         let pending_buffer_reply: Arc<Mutex<Option<tokio::sync::oneshot::Sender<Vec<u8>>>>> =
             Arc::new(Mutex::new(None));
         let on_exit: ExitHookSlot = Arc::new(OnceLock::new());
@@ -109,6 +138,7 @@ impl DaemonClient {
             notification_tx.clone(),
             pending_buffer_reply.clone(),
             on_exit.clone(),
+            for_recovery,
         )
         .await?;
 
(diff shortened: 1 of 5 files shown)


armaxri:
Holding merge on the #2490 Windows connect-flake fix (this PR rides the agent-integration suite, whose Windows leg is currently 10060-flaky). Once #2490 lands, rebase onto develop + re-run → Windows green → merge. This PR carries the **real reconnect fix** (recovery 2s connect-timeout for dead-lingering daemon sockets) + headless automated coverage; #2489 (agent-side coverage) also rebases after #2490.

Issue #2512, opened 2026-08-09
agent reconnect must re-attach the LIVE session (continue the running process), with an explicit session-lost state when unrecoverable


armaxri:
## The real requirement (maintainer, correcting the prior mis-spec)
When a resilient agent-hosted shell tab loses its transport and reconnects, the tab must **re-attach to the SAME live shell instance** — the shell process (and whatever it is running, e.g. a compile) stays alive across the disconnect and the reattach simply continues it. Example: start a compile, lose the connection, reattach → the compile is **still running** and continuing. NOT a new shell; NOT merely replayed scrollback.

**This supersedes the "reconnect with a NEW `sessionId`" wording in #2476/#2446/#2455/#2472/#2473** — that model (`create_connection` on reconnect) is exactly the wrong behavior and is what this issue replaces on the recovered-session path.

## Ground truth (live grade, 2026-08-09)
The **agent side already preserves the live process**: after a real transport drop+restore, the agent logs `Recovered session <id> (type=local)` / `Recovered 2 session(s)` — the setsid'd session daemons survive with their shell processes alive. The **desktop then wrongly calls `create_connection`** (`Created session <new-id>` + `Daemon spawned…`), orphaning the recovered live session. Fixing the drop harness (#2508/#2510) and the reconnect crash (#2503) was prerequisite to seeing this cleanly.

## The gap is desktop-side wiring only (~80% already exists)
- Agent RPCs already support attach-existing: `connection.list` (returns recovered sessions incl. `session_id`/`status`/`attached`), `connection.attach`, `session.getBuffer` (`agent/src/handler/dispatch.rs:635/682`). Recovery: `agent/src/session/manager.rs:856 recover_sessions()`, `:822 attach()`. Local shells are daemon-backed (`core/src/backends/local_shell.rs:389 persistent:true`).
- Desktop already has the attach-existing path: `RemoteProxy::reconnect_existing` (`remote_proxy.rs:104`) → `attach_session` → `connection.attach`, driven today by `attach_persistent_tab` (`session/manager.rs:1483`). Clients: `list_sessions`/`attach_session`/`get_remote_session_buffer`.
- **Missing:** the resilient redrive calls `create_connection` (`session/manager.rs:755`, via `redrive.rs:150`) instead of attaching, because the tab's agent `remote_session_id` is **not retained across the drop** (`RetainedConnectionRequest` carries `type_id`/`settings`/`agent_id`/`resilient`/`backend_reattach` but not the agent session id).

[...]

PR #2520, opened 2026-08-09
test(system): grade live agent shell continuity across backend-driven reattach (Part of #2512, #2480)

armaxri:
Extends the full-app Python system test for the live agent-reattach feature so it asserts the **real requirement** end to end: with `sessionBackendReattach` on, a dropped-and-restored agent transport must re-attach the tab to the **same live shell instance**, not spawn a fresh one.

Because the agent session daemon detaches from the sshd session (`setsid`), killing the harness sshd severs only the *transport* — the daemon-held shell keeps running — which is exactly what makes "the live process continues" testable through the real app.

## New backend-driven assertions (flag on)
- **Variable persists** — set `MYVAR` before the drop; after re-attach `echo`ing it still yields the value (a fresh shell would echo empty).
- **Running loop continues** — a self-incrementing loop's counter is past its pre-drop value after re-attach and keeps advancing (never paused or restarted at 0).
- **No duplicate scrollback** — a unique pre-drop marker still appears exactly once after re-attach (end-to-end guard for the #2515 / #2518 duplicate-scrollback regression). The marker is emitted so its *output* is the needle while the typed command text is not, so the buffer count reflects real emissions only.
- Keeps the existing basic drop → reconnect → re-attach (tab recovers live, not left exited) and the client-driven (flag off) recovery assertions.

## Un-gated
Removes the `TERMIHUB_TEST_LOCAL_AGENT_RECONNECT=1` opt-in so the suite runs directly in a foreground display session; it still skips cleanly when `sshd` or the agent binary is absent.

## Verification
Collectable and correct by construction — `cd tests/system && ./pytest.sh --collect-only -q -k AgentReconnect` lists all 5 tests with no import/syntax errors. The live green grade needs a real display session (a headless WKWebView occlusion-throttles the long connect, #2480), so it is a coordinator/manual run.

Run the extended tests with:
`cd tests/system && ./pytest.sh -m integration -k "AgentReconnectBackendDriven" -v -s`

Part of #2512
Part of #2480

### DIFF
--- a/tests/system/tests/test_agent_reconnect_live.py
+++ b/tests/system/tests/test_agent_reconnect_live.py
@@ -1,4 +1,4 @@
-"""Live agent drop / reconnect over a harness-controlled sshd (#2476).
+"""Live agent drop / reconnect over a harness-controlled sshd (#2476, #2512, #2480).
 
 The missing enabler for the agent reconnect activation: no scenario existed to
 drive a real agent-transport drop and assert the reconnect outcome. The desktop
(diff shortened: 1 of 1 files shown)


armaxri:
Superseding with a log-based full-app test: the webview-bridge assertions can't read state through the WKWebView occlusion-throttle during the long reconnect. The replacement asserts reconnect+recovery via the app's backend log (throttle-immune). Scenarios (var/loop/no-dup) preserved; content-continuity is covered headlessly by the #2519 agent-integration tests.

PR #2522, opened 2026-08-09
test(system): log-based full-app agent-reconnect suite (Part of #2480, #2512)

armaxri:
## What

Rewrites the full-app system test `tests/system/tests/test_agent_reconnect_live.py` to be **log-based** so it runs green in a display session **without depending on webview-bridge reads during the reconnect**.

## Why

During the long agent reconnect the WKWebView occlusion-throttles, so polling the frontend state via the bridge times out (`command timed out after 60s`) even though the app reconnects fine. This **supersedes the bridge-poll approach of #2520** for exactly that reason. The app's **backend log is written by Rust and is immune to that throttle**, so this suite asserts on it via the harness's per-instance capture, `AppInstance.read_log()`.

## How

- **Setup stays on the bridge (pre-drop, un-throttled):** seed + connect the agent, open a shell, confirm the initial `connected` state and a live round-trip.
- **After the drop+restore, assert on the LOG, not the bridge** (bounded polls on `app.read_log()`, no fixed sleeps):
  - drop surfaces: `connection lost, attempting reconnect`
  - reconnect handshake completes again: `initialize response received (...); marking connected`
  - **flag on (backend-driven):** the live daemon session is re-attached for the same `<sid>` — `Recovered session <sid>` / `Reattached to session <sid>` (agent stderr forwarded into the desktop log) — and **no** fresh `Daemon spawned for session` (== a new `Created session`) and **no** `Failed to recover session` (== the `session_lost` fold) appear post-drop.
  - **flag off (client-driven):** log-based "surfaces the drop and recovers" — the single-shot transport reconnect gives up (`reconnection failed`) while the server is down, then the handshake completes again after the server returns and the connection is re-initiated.
- **Terminal *content* continuity is deliberately not read back through the throttled bridge** — it is covered headlessly by the agent-integration tests (`fresh_agent_recovers_daemon_session_from_dead_prior_agent`, the #2519 tests).
- **Un-gated:** dropped the `TERMIHUB_TEST_LOCAL_AGENT_RECONNECT=1` opt-in so it runs in a display session; still skips cleanly when sshd / the agent binary is absent.

## Verification

Cannot run green here (needs a real display — run via `launchctl asuser`). Collects cleanly:

```
cd tests/system && ./pytest.sh --collect-only -q -k AgentReconnect
# 2/642 tests collected
```

Part of #2480, #2512

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -125,6 +125,9 @@ objc2-foundation = { version = "0.3", features = [
     "NSUserDefaults",
     "NSString",
     "NSArray",
+    # Hold an NSProcessInfo activity assertion under the test bridge so macOS
+    # App Nap never throttles the headless E2E app (see utils::macos_unthrottle).
+    "NSProcessInfo",
 ] }
 # Native Cocoa Services provider for the app-level "Open in termiHub"
 # Services-menu entry (#1409): register on NSApp.servicesProvider and read the
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -279,6 +279,18 @@ pub fn run() {
         spawn::Command::None => None,
     };
 
+    // Under the headless test bridge on macOS, turn off AppKit window occlusion
+    // detection *before* NSApplication launches (it reads this user default
+    // while launching). Without it, macOS marks the unfocused test window
+    // occluded and WebKit throttles the page's timers, stalling the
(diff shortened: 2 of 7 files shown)


armaxri:
Blocked on **#2526** (display-backed E2E runner). This log-based full-app test is correct and connects, but the frontend reconnect engine won't advance headlessly without an actively-composited display (11 live runs confirmed). Hold this PR until #2526 lands, then un-skip.

